PR #98, opened 2017-04-30
Gaucho 0.2.2

angrykoala:
* Tasks with same name will now append the number between parenthesis
  * Fixed bug where editing and saving tasks with the same name will append 2
  * Small improvements in unit tests
  * jshint code style checks added as part of the tests
  * Minor code and style refactoring


angrykoala:
Electron is having problems instaling (https://github.com/electron/electron/issues/9323) so gaucho cannot be safely released yet

PR #131, opened 2017-08-27
Dev

grg121:
Bottom bar added and working more or less..

When the app starts, totalTaks is set to 0. It is actualiced when switching to other suite but I haven't been able to made it initialiced at start :P


app.js
```diff
@@ -8,7 +8,8 @@ const Material = require('./app/renderer/materialize');
 const components = {
     "task-suite": require('./app/renderer/components/task_suite'),
     "navbar": require('./app/renderer/components/navbar'),
-    "about": require('./app/renderer/components/about')
+    "about": require('./app/renderer/components/about'),
+    "bottombar": require('./app/renderer/components/bottombar'),
```
angrykoala: By convention in gaucho/css, componentes names should be separate using `-`, in this case, `bottombar` should be `bottom-bar`

angrykoala: The file name should be renamed to `bottom_bar.js` as well please

app/renderer/components/bottombar.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+const AppStatus = require('../app_status');
+
+module.exports = {
+
```
angrykoala: Unnecessary white line

app/renderer/components/bottombar.js
```diff
@@ -0,0 +1,27 @@
+"use strict";
+
+const AppStatus = require('../app_status');
+
+module.exports = {
+    
+    props: ['suite'],
```
angrykoala: Is the suite prop used in the component.

In vue.js props are the standard way of passing data from parent component to child component, you should only pass data that you actually need for rendering

app/renderer/components/bottombar.js
```diff
@@ -0,0 +1,27 @@
...
+        <nav class="navbar navbar-inverse">
+             <div id="left">
+                <b>Total Tasks:</b> {{AppStatus.totalTasks}}
+            </div>
+            <div id="right">
+                <b>Running Tasks:</b> {{AppStatus.runningTasks}}
+            </div>
+    </nav> 
+    </div>
+    `,
```
angrykoala: Unnecessary trailing comma and white line

app/renderer/components/navbar.js
```diff
@@ -66,6 +66,7 @@ module.exports = {
         },
         onTabSelected(index) {
             AppStatus.activeSuite = index;
+            AppStatus.totalTasks = this.suites[index].length;
```
angrykoala: With this, I understand you only count the tasks of the current suite. Should total tasks count every task in gaucho?

(this can easily be changed here with a [reduce](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) function

angrykoala: And the value wouldn't change on each tab. We should discuss the exact use case of the variable `totalTasks` on an issue

app/renderer/components/task_card.js
```diff
@@ -81,7 +81,12 @@ module.exports = {
         },
         run() {
             this.output = "";
-            this.task.run(this.print, () => {});
+            AppStatus.runningTasks++;
+            this.task.run(this.print, () => {   AppStatus.runningTasks--;   }
```
angrykoala: Style note: the function should be multilined. 
While in this project is not yet defined and standardized (my fault) all the gaucho JS code is usually passed through a beautifier (https://www.npmjs.com/package/js-beautify) to generate a consistent style, if you are not sure of the coding style followed in this projet, feel free to try and execute that beautifier to automatically comply with the convetions

(Also, I should add a new issue to add the beautify as an npm script, to make this easier)

index.html
```diff
@@ -3,7 +3,7 @@
 
 <head>
     <meta charset="UTF-8">
-    <title>Gaucho Task Launcher</title>
+    <title>Gaucho Task Runner</title>
```
angrykoala: Why this change in the title?

The title was change on purpose to Launcher, as I think it defines better what gaucho is for.

If you think another title/name is better, feel free to open an issue so it can be deiscussed

index.html
```diff
@@ -12,13 +12,24 @@
 </head>
 
 <body>
-    <div id="app">
+<div id="app">
```
angrykoala: Again, html files follows a convention that is better to be followed (tabs and spaces..), you can automatically "fix" that with js-beautify, like before

index.html
```diff
@@ -12,13 +12,24 @@
...
         <about></about>
-        <navbar v-bind:suites="suites" v-if="loaded"></navbar>
-        <template v-for="(suite,index) in suites">
-            <task-suite v-bind:suite="suite" v-bind:index="index"></task-suite>
-        </template>
-    </div>
+        <div class="r_flex_container">
```
angrykoala: what does `r flex container` means? if possible, add more intuitive names to css classes

index.html
```diff
@@ -12,13 +12,24 @@
...
+            <div class="r_flex_fixed_child_top">
```
angrykoala: By convention, all css classes and ids should be separated using `-`

For example: `r-flex-fixed-child-top`

index.html
```diff
@@ -12,13 +12,24 @@
...
+                <navbar v-bind:suites="suites" v-if="loaded"></navbar>
+            </div>
+            
+            <div class="r_flex_expand_child">
```
angrykoala: are these divs needed out of components? If possible, add them as part of the component or as separate components.

If it is too hard, complex, or impossible, just leave them as they are, It is not a big problem

resources/styles.css
```diff
@@ -30,14 +30,21 @@
     background-color: #af0000;
 }
 
+.collapsible{
+ margin-top: 0px;
```
angrykoala: Per css convention, a 0 value should not have type (`px`), so just write `margin-top: 0;`

resources/styles.css
```diff
@@ -30,14 +30,21 @@
...
+ border-top: 0px;
+}
+
+
 .collapsible-body.task-card-body {
     padding: 0;
+
```
angrykoala: Why a new white line?

resources/styles.css
```diff
@@ -69,7 +76,8 @@
 }
 
 .suite-tab {
-    margin-top: 47px;
+    margin-top: 0px;
```
angrykoala: Nice that we don't need the magic number in suite-tab and run-card anymore!  (they where a lot of problems)

Please, try to directly remove the css got suite-tab, run-card padding and border, and see if everything is still correct (we may not need these styles anymore)

resources/styles.css
```diff
@@ -211,3 +219,50 @@ input.tab-text {
     width: 20px;
     height: 20px;
 }
+
+
+
+
+html, body, .r_flex_container, #app{
```
angrykoala: using html and body is not recommended, do we really need that?

resources/styles.css
```diff
@@ -211,3 +219,50 @@ input.tab-text {
...
+    flex:none;
+    margin-bottom: 0px;
+}
+
+.r_flex_expand_child {
+    flex:auto;
+    overflow-y:auto;
+}
+
+#left {
```
angrykoala: left and right should **not** be ids.

css ids should be unique, this means that they should be avoided for things that we are not sure will be unique (like right elements)

Also, make sure there is not any unwanted collision with these, as left/right are extremely common css classes

resources/styles.css
```diff
@@ -211,3 +219,50 @@ input.tab-text {
...
+  padding-left: 20px;
+    
+}
+#right {
+  padding-right: 20px;
+  width: 100%;
+  text-align: right;
+}
+
+ul{
```
angrykoala: Why do we need this?

app/renderer/components/navbar.js
grg121: discuss: https://github.com/angrykoala/gaucho/issues/132

index.html
grg121: I don't remember have made this change (?) maybe I just copy it from the main branch or something XD, sorry.

index.html
grg121: You're right. flex container was because the property flex that allow me to fix the space between lasts visible tasks and bottom bar, and r was only  to differenciate it. 
I'll change names to made them more intuitive

index.html
grg121: I think there are neededs D:

resources/styles.css
grg121: fail :D

resources/styles.css
grg121: perfect :P

resources/styles.css
grg121: Actually not. I saw this on an example and just do it but I'm sure we could have the same effect in other way... I'm going to try to change it

grg121: well... I'm not sure why but I can get the same effect without this XD 
I think what is applied to html and body:
  height: 100%;
    display: flex;
    flex-direction: column;
    margin: 0;

is inoffensive and (from now) necessary :/

resources/styles.css
grg121: Ops, I did that just trying and forgot to change it! I'll do

resources/styles.css
grg121: It seems there are some css files at some node_modules and one of them have rules to <ul> elements that made them have a top-margin which made the problem with right scroll bar. 
I'm not sure why but it seems that your styless.css has preference to these css files and, as I didn't know if  we could change them or if it would be a good idea, I decided to try writting this rule and it did the trick.
Then I realiced that, as the list that was generating troubles has the class collapsible, it would be the same if I put that rule at collapsible (what I did) but I forgot to remove it from ul.

The file I mentioned is.. materialize.min.css, which made a top margin of 7px

So.. I remove that rule!

grg121:
Now I'm gonna do the part that should initialize total-tasks at the start and, when we decided what exactly this variable should represent, we could just change it if needed.

index.html
angrykoala: No Problem, thats what Code reviews are for :+1:

grg121:
Now, bottom bar seems to work fine :P

angrykoala:
The new bottom bar seems to be working fine and the new way scrolling works is great. However, the style of the task cards have changed (probably some change messing)

screenshot from 2017-09-20 23-20-14
_Previous version_
screenshot from 2017-09-20 23-20-31
_Current one_
 with the css) as you can see, the new version is more "cluttered"

I will investigate a bit to see if I can find where the problem is, if you have an idea, please comment

angrykoala:
Please, avoid commiting project and IDE related files (.vscode)

Issue #140, opened 2017-09-28
Bottom bar text shouldn't be selectable


angrykoala:
Add the needed css classes

PR #142, opened 2017-09-28
Fix: Bottom bar text shouldn't be selectable.  #140

sergeir82:
Add default cursor to unselectable-text class.


angrykoala:
@sergeir82 Everything good, but please, do your PR against the `dev` branch

Issue #141, opened 2017-09-28
Avoid rendering Config menu when not needed


angrykoala:
Configuration menu is always been rendered. Try adding an `v-if` so it is not rendered until needed

angrykoala:
Solving this is more complicated that I expected, as adding an v-if will mess with the materialize animations, as proven by @panvourtsis in #144

PR #144, opened 2017-09-28
Avoid rendering Config menu when not needed

panvourtsis:
this PR is fixing the #141

### DIFF
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -9,7 +9,6 @@ const defaultData = {
     outputMaxSize: 10000,
     maximized: false,
     bottomBar: true,
-    configMenu: false,
     animatedSpinner: true,
 };
 
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -9,6 +9,7 @@ const defaultData = {
     outputMaxSize: 10000,
     maximized: false,
     bottomBar: true,
+    configMenu: false,
     animatedSpinner: true,
 };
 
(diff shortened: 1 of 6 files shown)


angrykoala:
Thanks for contributing.
This solution seems right, but I'm not sure why are you storing the flag `configMenu` in the actual app configuration (which will be stored). This could be simply stored in AppStatus (no data from here is actually stored anywhere), defaulting to false, the remaining logic should be the same

panvourtsis:
i though you needed for some reason based on your code and description. I can simply store it, shall i?

angrykoala:
Yes, I'm sorry this stuff is not very well documented (just yesterday I refactored the config)

In a nutshell, AppStatus is simply a way of stored global variables (like flags to render config menu), while app_config is going to actually be stored across sessions.

A variable on AppStatus (or any other place) should be enough

angrykoala:
* I don't think we need this variable as part of the config in config menu. It should be enough with having it in appstatus
* Please, make the pr to dev :)

panvourtsis:
you are right, getting familiar with you code now

angrykoala:
Don't worry that's what code reviews are for

angrykoala:
After testing your version, I found a problem with doing this. When closing the config menu, it closes suddenly (because it stops being rendered). I think my idea of stop rendering config is not as simple as I though. Do you have any idea on how to improve this?

panvourtsis:
I've seen the same while i was doing it. Vue offers this modal but even if you load it with template inside the page the script will be there. Even on online demos i can see that. What i've seen working is [here](https://www.npmjs.com/package/vue-js-modal) demos are [here](http://vue-js-modal.yev.io/)

Let me know if these are working in your case though

angrykoala:
To be honest, I don't think is worth it adding another library for this, the performance impact of config is not that big. I'm afraid I'll have to close this PR, but thanks for the investigation, I'll close the issue as it is not worth it for something as simple as the config

Issue #145, opened 2017-09-28
introduce pre commit functionality


panvourtsis:
Can we have a pre commit functionality to run necessary procedures like jshint for example before letting anyone commit. I found this very useful

angrykoala:
Wouldn't it be a hassle to run all those checks with every commit?, I only care of having everything right at the moment of creating a pull request and all checks can be done with `npm test` (including jshint).

Anyway, I'm ok with adding the pre-commit script pointing to npm test to see how it works

panvourtsis:
avoiding though any additional commits to fix the jshinting etc. will create a pr later in the day for you to see

angrykoala:
@panvourtsis could you describe how to make the pre-commit functionality work? after fresh clone and `npm install`, the pre-hooks are still the default ones and no checks are done before commit

panvourtsis:
hi @angrykoala i just checked it on a new branch and is all good. Maybe the pre-commit wasnt installed?

angrykoala:
Have you tried from a fresh clone? tbh I'm not sure how hooks work. `npm install` installed the pre-commit package, but it seems to ignore the actual hook setup

panvourtsis:
yes i did try from a fresh clone. Actually my test is as follows

1. clone
2. checkout to the dev
3. change the task.js by adding a second semicolon so the jshint to got an error
4. trying to commit

you can try an error on the tests as well you can see that it will not commit it.

angrykoala:
Thanks for the specific steps, I'll try it as soon as I can, and come back to you with the result

angrykoala:
Indeed, it seems to be working fine, probably a mees-up with my env. I'll close this issue then
Thanks

PR #146, opened 2017-09-28
introduce pre commit

panvourtsis:
what we discussed on #145

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "jshint": "jshint *.js app test"
   },
   "pre-commit": [
-    "jshint"
+    "test"
   ],
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -18,11 +18,12 @@
     "istanbul": "^0.4.5",
     "jshint": "^2.9.4",
     "mocha": "^3.2.0",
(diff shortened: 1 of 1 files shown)


angrykoala:
pre-commit should be "test" instead of jshint, so it runs both, the tests and jshint

panvourtsis:
do you have to run clean everytime for testing?

angrykoala:
Not really, Maybe it would be better removing the clean and jshint from npm test and add those to your precommit. What do you think?

panvourtsis:
on my projects i have two scripts runnning on precommit the tests and the linting. In your case 

`["_mocha", "jshint"]`

angrykoala:
the tests also use istanbul, wouldn't it be better to execute the npm scripts I already have?
```
scripts:{
  "test": "istanbul cover _mocha",
  "clean": ...,
  "jshint": ...
},
precommit:[ "test","clean","jshint"]
```

Does this make sense?

panvourtsis:
this is what i use (without the clean)

angrykoala:
Yeah, I'll give it a shot and see if it makes easier contributions. Please, point the PR to the dev branch

Issue #87, opened 2017-04-08
Support rearranging tasks


angrykoala:
It should be possible to rearrange tasks, so the order of the tabs can be changed or moved to a different suite

panvourtsis:
thinking about something like sortablejs ?

angrykoala:
The main difficulty here is making whatever we use work with materializecss collapsibles (http://materializecss.com/collapsible.html), each task card should have a icon or something as the drag area to move the cards around.

If this is not possible (maybe collapsible makes it very hard or something) we could simply use a couple of arrow buttons to move the tasks when edit mode is on

panvourtsis:
i can start implementing the arrows and also investigate to see if its possible to d&d if not we will stick with arrows

angrykoala:
Looks like a plan, thanks!

The moving around may be a bit tricky, check the code for deleting tasks, as it contains all the logic to move tasks around as well

PR #149, opened 2017-09-29
initial commit on ordering

panvourtsis:
i introduced ordering on tasks based on their index, on change i am updating the saved ordering
I also installed the draggable for testing

need to figure out how to update on drag and drop. have a look as well

**this solution is about**
the issue #87

### DIFF
--- a/app/renderer/components/add_task.js
+++ b/app/renderer/components/add_task.js
@@ -3,11 +3,13 @@
 const TaskInput = require('./task_input');
 
 module.exports = {
+    props: ['tasks'],
     data() {
         return {
             title: "",
             command: "",
             path: "",
+            order: 0,
         };
     },
     template: `
@@ -19,7 +21,7 @@ module.exports = {
             </strong>
         </div>
         <div class="collapsible-body container task-card-body">
-            <task-input v-on:save="addTask"></task-input>
+            <task-input v-bind:order="tasks.length + 1" v-on:save="addTask"></task-input>
         </div>
     </li>    
     `,
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -26,11 +26,15 @@ module.exports = {
     template: `
     <li class="run-card">
         <div class="collapsible-header row unselectable-text">
-            <div class="col s5">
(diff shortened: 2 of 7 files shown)


angrykoala:
Hi, thanks for working on this so quickly
A few comments, I hope this will help you with this issue:

* I don't think we need an `order` value in the task, the tasks are rendered from an array (`[task1, task2]`) so the order is already implicit in that array, the arrows should reorder within the array.

* While testing your current version, I saw that the feature of displaying the stdout when clicking a task is broken (I assume is because your version is incomplete), just pointing out so you are aware of that problem

* I assume the task.order you are displaying is for testing purposes

* As your arrows are being displayed only when edit mode is active, like the delete button, I recommend adding all of them to a single div with 1 v-if

* Also, I think the down arrow should be below the up arrow

* changeItemOrder requires ev.stopPropagation()?

* Don't worry about the draggable for now, I'm happy to accept a PR with the arrow buttons and create a new issue to improve the draggable behavior, one thing at a time.

* Most things are quite simple to fix, the only important problem I see is the use of ordering, I'm quite sure that it will be simpler to use the array implicit order, my recommendation:
  * Don't make any modifications to the task (only change task-card to add the arrows)
  * The behaviour of reordering tasks should rely completely on the task suite
  * Check the `delete task` code, as the problem is quite similar, as it required to reorder the tasks below the deleted task

I'm not sure if you have any other question, feel free to ask. I will close this PR

Good job, and thanks for your help

panvourtsis:
Hey, order value is just for testing purposes there. 

What i was thinking is to have the arrows and each task will have initially their index values. This eventually will be saved on task thats why you need it on the constructor. am i right?

Removing a task is not that a big a deal as moving something. Not as much as moving it 1 by 1 but on drag and drop when you might move it 3 or 4 lines at once.

angrykoala:
Ok, I didnt know it was for testing. The final tasks config should not keed any change, as I said, the tasks being in an array already provide the order

I know it is not the same, but it was a bit of already made code to help you get started. If you are able to cover all cases now, good, if not, It's enough to cover just the move one position for now until we implement the drag and drop :)

Issue #23, opened 2016-11-28
Export tasks


angrykoala:
Export the tasks.json (as a gtask.json file or something like that) to be able to import/load into a different gaucho instance ( #20 )

PR #150, opened 2017-09-30
solve #23

grg121:
Maybe would be nice to set a default save directory but I'm not sure how do it and what directory should we use for this (depend on the user's OS) so I have not set it...

Seems to work, if you notice something just tell me :)

### DIFF
--- a/app.js
+++ b/app.js
@@ -21,8 +21,8 @@ ipcRenderer.on('before-close', () => {
     const promises = suites.map((s) => {
         return s.stopAll();
     });
-    Promise.all(promises).then(() => {
-        ipcRenderer.send("close-app");
+  Promise.all(promises).then(() => {
+    ipcRenderer.send("close-app");
     });
 });
 
--- a/app/renderer/components/config_menu.js
+++ b/app/renderer/components/config_menu.js
@@ -1,4 +1,8 @@
 "use strict";
+
+const os = require('os');
+const path = require('app');
+
 const app = require('electron').remote;
 const dialog = app.dialog;
 const fs = require('fs');
@@ -51,21 +55,20 @@ module.exports = {
       let content = TaskConfig.getData() ;
       content = {
         "suites":content,
-        "version":"0.3.4"
+        "version":AppStatus.version
       };
       content = JSON.stringify(content) ;
-      dialog.showSaveDialog(
-        {
-          defaultPath: require('os').homedir() + "/gtask.json",
-          filters: [
-            { name: 'json', extensions: ['json'] }
-          ]}, (filename)=> {
-            if(filename){
-              fs.writeFile(filename, content, (err) => {
-                if(err) console.warn(err);
-              });
-            }
-          });
+      dialog.showSaveDialog({
+        defaultPath: path.join(os.homedir(),"gtask.json"),
+        filters: [
+          { name: 'json', extensions: ['json'] }
+        ]}, (filename)=> {
+          if(filename){
+            fs.writeFile(filename, content, (err) => {
+              if(err) console.warn(err);
+            });
+          }
+        });
     },
     clearTasks() {
       TaskConfig.clearTasks();
--- a/app/renderer/components/config_menu.js
+++ b/app/renderer/components/config_menu.js
@@ -13,12 +13,12 @@ module.exports = {
         return {
             config: {
                 bottomBar: AppStatus.config.bottomBar,
-                animatedSpinner: AppStatus.config.animatedSpinner,
+                animatedSpinner: AppStatus.config.animatedSpinner
             }
         };
     },
     components: {
-        "switch-form": SwitchForm,
+        "switch-form": SwitchForm
     },
     template: `
     <div id="config-modal" class="modal bottom-sheet modal-fixed-footer">
@@ -53,23 +53,20 @@ module.exports = {
         "suites":content,
         "version":"0.3.4"
       };
-
       content = JSON.stringify(content) ;
-
-        dialog.showSaveDialog({ filters: [
-
-          { name: 'json', extensions: ['json'] }
-
-        ]}, (filename)=> {
-
-          if(filename){
-                     fs.writeFile(filename, content, (err) => {
-                       if(err) console.warn(err);
-                     });
-          }
-        });
+      dialog.showSaveDialog(
+        {
+          defaultPath: require('os').homedir() + "/gtask.json",
+          filters: [
+            { name: 'json', extensions: ['json'] }
+          ]}, (filename)=> {
+            if(filename){
+              fs.writeFile(filename, content, (err) => {
+                if(err) console.warn(err);
+              });
+            }
+          });
     },
-
     clearTasks() {
       TaskConfig.clearTasks();
       AppStatus.activeSuite = 0;
--- a/app/renderer/components/config_menu.js
+++ b/app/renderer/components/config_menu.js
@@ -1,8 +1,12 @@
 "use strict";
+var app = require('electron').remote;
+var dialog = app.dialog;
+const fs = require('fs');
 
 const AppStatus = require('../app_status');
 const SwitchForm = require('./switch_form');
 const TaskConfig = require('../task_config');
+
 
 
 module.exports = {
@@ -29,6 +33,10 @@ module.exports = {
                     <a class="waves-effect waves-light btn modal-action modal-close " v-on:click="clearTasks">Clear Tasks</a>
                     <label>Warning: This will remove all your suites and tasks</label>
                     <a class="waves-effect waves-light btn" v-on:click="resetConfig">Reset Configuration</a>
+                     </br>
+                     <a class="waves-effect waves-light btn" v-on:click="exportTasks">Export Tasks</a>
+                    <label>Export the tasks.json to be able to load it into a different gaucho instance
+                    </label>
                 </div>
             </div>
         </div>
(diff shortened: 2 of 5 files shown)


angrykoala:
Please, do the PR to dev branch, it is hard to make the code review with changes other than yours :)

grg121:
ups, sorry.
Done.

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,41 @@ module.exports = {
...
-        onClose() {
-            this.config.bottomBar = AppStatus.config.bottomBar;
-            this.config.animatedSpinner = AppStatus.config.animatedSpinner;
-        },
-        onSave() {
-            AppStatus.config.bottomBar = this.config.bottomBar;
-            AppStatus.config.animatedSpinner = this.config.animatedSpinner;
-        }
+  methods: {
+
```
angrykoala: Please, remove unnecessary empty lines

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,41 @@ module.exports = {
...
+    exportTasks(){
+      let content = TaskConfig.getData() ;
+      content = JSON.stringify(content) ;
+
+        dialog.showSaveDialog({ filters: [
+
+          { name: 'gtasks', extensions: ['json'] }
+
+        ]}, function (fileName) {
```
angrykoala: Use an arrow function here (and change fileName to filename):

(filename)=>{
....

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,41 @@ module.exports = {
...
+
+          if(fileName === undefined) return ;
+
+          fs.writeFile(fileName, content, (err) => {
+            if(err) console.log(err);
```
angrykoala: Use `console.warn(err)`

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,41 @@ module.exports = {
...
+      let content = TaskConfig.getData() ;
+      content = JSON.stringify(content) ;
+
+        dialog.showSaveDialog({ filters: [
+
+          { name: 'gtasks', extensions: ['json'] }
+
+        ]}, function (fileName) {
+
+          if(fileName === undefined) return ;
```
angrykoala: What happens if fileName is an empty string?

I'm not sure, but maybe `if(!filename)` is a better check.

Also, instead of return ;, do the folowing:
if(filename){ //or filename !== undefined
//rest of the code
}

app/renderer/task_config.js
```diff
@@ -9,74 +9,83 @@ const key = "Z4xu6Nzj";
 
 
 const defaultTasks = [{
```
angrykoala: I think you did a beautify in this file with tab=2 spaces, in this project we are using tab=4 spaces

angrykoala: Because of this I'm not completely sure if you did another change to this file

app/renderer/components/config_menu.js
grg121: by default, download dialog ask you to put a file name if you let an empty string. I've seen various examples where people onli check if the filename is undefined (I'm not sure how can you set filename to undefined but...) 
If you prefer, I'll put !filename :P

app/renderer/task_config.js
grg121: Um... I think it was emacs... I've a plugin that automatic indent the code if selec all and push tab and maybe I've done that by mistake.

I didn't change nothing here... sorry D:

grg121: Not sure if I've propertly removed this file from the commit D:

angrykoala:
screenshot from 2017-10-03 09-30-03

Is it working in your machine? Probably just a typo or something outdated in `app.js`

angrykoala:
* Missing default filename. When exporting, a default name should be present in the dialog
* After thinking a bit, I think the default format should be `.json` instead of `.gtasks` (sorry for confusion in the issue)

* the json format looks good. But I think that instead of directly sending the array, it would be better to send an object with a "suites" property. This makes explicit what the file contains and allows for future expansion
```json
{
"suites": [ ...(the current file) ...]
}
```

In fact, I would say about adding the current Gaucho version to the file:
```json
{
"suites": [....],
"version": "0.3.4"
}
```
This way, we can handle compatibility issues in the future. (**tip** the version can be found on AppStatus.version)

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,40 @@ module.exports = {
...
-            AppStatus.config.bottomBar = this.config.bottomBar;
-            AppStatus.config.animatedSpinner = this.config.animatedSpinner;
-        }
+  methods: {
+    exportTasks(){
+      let content = TaskConfig.getData() ;
+      content = JSON.stringify(content) ;
+
+        dialog.showSaveDialog({ filters: [
+
```
angrykoala: Please, remove unnecessary spacing and new lines, e.g.:
```js
filters:[
{ name: 'gtasks', extension ['json'] }
]},
```

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,40 @@ module.exports = {
...
+          { name: 'gtasks', extensions: ['json'] }
+
+        ]}, (filename)=> {
+
+          if(filename){
+                     fs.writeFile(filename, content, (err) => {
```
angrykoala: Aaand, use a 4 spaces tab please

app/renderer/components/config_menu.js
```diff
@@ -1,4 +1,7 @@
 "use strict";
+const app = require('electron').remote;
+const dialog = app.dialog;
```
angrykoala: as a note for the future (**no need to change this now**) personally I like to add the main dependencies first, and any change at the bottom. These kind of agreements are simply for consistency and ease of reading.

So, here it would be (comments added to explain):
```js
const fs = require('fs'); // System built-in library, no dependency
const app = require('electron').remote; // External dependency
              // Space to show the difference between dependencies and our code
const AppStatus = require('../app_status');  // These are our own modules
const SwitchForm = require('./switch_form');
      // again, space to show the following are not imports
const dialog = app.dialog; // constant, not a dependency, but a variable/process
```
Each project has its own conventions on this, there is not a strict rule, I simply try to follow the convention explained in the example in this project

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
-        onSave() {
-            AppStatus.config.bottomBar = this.config.bottomBar;
-            AppStatus.config.animatedSpinner = this.config.animatedSpinner;
-        }
+  methods: {
+    exportTasks(){
+      let content = TaskConfig.getData() ;
+      content = {
+        "suites":content,
+        "version":"0.3.4"
```
angrykoala: Please, don't hardcode versions, use `AppStatus.version`

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
-            this.config.bottomBar = AppStatus.config.bottomBar;
-            this.config.animatedSpinner = AppStatus.config.animatedSpinner;
-        },
-        onSave() {
-            AppStatus.config.bottomBar = this.config.bottomBar;
-            AppStatus.config.animatedSpinner = this.config.animatedSpinner;
-        }
+  methods: {
+    exportTasks(){
+      let content = TaskConfig.getData() ;
```
angrykoala: It is a good idea to reuse the same variable so many times? maybe the process of parsing the content into a JSON could be a private method?

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
+      content = {
+        "suites":content,
+        "version":"0.3.4"
+      };
+      content = JSON.stringify(content) ;
+      dialog.showSaveDialog(
+        {
```
angrykoala: please, remove unnecessary lines, you can write `dialog.showSaveDialog({`

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
+          defaultPath: require('os').homedir() + "/gtask.json",
```
angrykoala: Few things:
1. using require in the middle of the code is evil, as is hard to spot dependencies, do `const os=require('os')` at the beginning
2. pro node.js tip: You code here is correct, but node provides a built-in module to handle paths, so instead of concatenating strings, using this library you guarantee the path is a valid one (e.g. backslash in windows):
```
const path=require('path');
const os=requie('os');
// rest of the code
defaultPath: path.join(os.homedir(),"gtask.json")

File: app/renderer/components/config_menu.js [hunk 13]
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
+          filters: [
+            { name: 'json', extensions: ['json'] }
```
angrykoala: json.json? to be honest I don't understand this "filters" bit

app/renderer/components/config_menu.js
```diff
@@ -39,22 +46,42 @@ module.exports = {
...
+          ]}, (filename)=> {
+            if(filename){
+              fs.writeFile(filename, content, (err) => {
```
angrykoala: aaand what happens if no filename? and why wouldn't be filename?

If there is a chance that the filename doesn't exists (i guess it is because the user didn't set a path or something) it doesn't makes sense all the processing we did to create the content.

I would say it is better to create the content just before writing it. Unless there is a reason not to do it that way

app/renderer/components/config_menu.js
grg121: not actually... name is not exactly the file name. 
ss_20171008_131533

Look at the img... the name parameter change not the file name but the "extension" name that appears next to the save buttom. (I'm not sure what it could be used for) 

I've tried to skip it and by default take the same value as extensions so we could just remove it... other option would be set it to "gtasks-json" to inform of what this JSON is for but I'm not sure if it would be confused.

app/renderer/components/config_menu.js
grg121: Actually, I would say that the user could not save the file without select a filename, but in some guides I read people check it so I did... maybe for something we're ommiting... so, I think we should let this check and just modify when the content is created as you said :D

app/renderer/components/config_menu.js
angrykoala: Ok, I understand, anyway the current behaviour looks fine to me, so Its ok to leave it at it is

app.js
```diff
@@ -21,8 +21,8 @@ ipcRenderer.on('before-close', () => {
     const promises = suites.map((s) => {
         return s.stopAll();
     });
-    Promise.all(promises).then(() => {
-        ipcRenderer.send("close-app");
+  Promise.all(promises).then(() => {
```
angrykoala: Why did this change?

PR #151, opened 2017-10-02
Ordering

panvourtsis:
## For which issue
This is targeting the #87 issue

## What i am doing
I introduce an order functionality across tasks which can be changed either by arrows or by drag and drop. There is a **stopPropagation** to avoid on click to call the expand of the collapsible. Parameter of order is being on task to save the order which depends on the index of the item + 1. I have also introduce a sorting by ordering on the tasks before listing them

## Not ready yet.
I need your help @angrykoala  to decide how the arrows should look

### DIFF
--- a/app/renderer/components/add_task.js
+++ b/app/renderer/components/add_task.js
@@ -3,11 +3,13 @@
 const TaskInput = require('./task_input');
 
 module.exports = {
+    props: ['tasks'],
     data() {
         return {
             title: "",
             command: "",
             path: "",
+            order: 0,
         };
     },
     template: `
@@ -19,7 +21,7 @@ module.exports = {
             </strong>
         </div>
         <div class="collapsible-body container task-card-body">
-            <task-input v-on:save="addTask"></task-input>
+            <task-input v-bind:order="tasks.length + 1" v-on:save="addTask"></task-input>
         </div>
     </li>    
     `,
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -26,11 +26,15 @@ module.exports = {
     template: `
     <li class="run-card">
         <div class="collapsible-header row unselectable-text">
-            <div class="col s5">
-                <strong class="truncate">{{task.title}}</strong>
+            <div class="col s4">
+                <strong class="truncate">{{task.title}} {{task.order}}</strong>
             </div>
-            <div class="col s3">
+            <div class="col s2">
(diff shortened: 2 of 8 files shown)


angrykoala:
@panvourtsis can you put a screenshot of how the arrows are looking now?

If the drag&drop is going to be working, we may not even need the arrows, what do you think?

For the drag and drop, the drag handle material icon may be appropriate
screenshot from 2017-10-02 14-28-20

Also, we could have this discussion in the issue itself, as it will be easier to find in the future :)

Thanks for the update!

panvourtsis:
screen shot 2017-10-02 at 15 30 57
I open it for you to test it as well. 

We can put the icon at the right end. I agree that arrows are not needed but many users like them i put them in because it's not that big deal for me to add or remove them.

Check the screenshots
screen shot 2017-10-02 at 15 31 04
screen shot 2017-10-02 at 15 30 57

angrykoala:
Well, when in edit mode the task cards should have the exact same size so the arrows are closer together.

Is the drag&drop working? if so, add the other icon on one of the sides to we can check if both are over cluttered. It is usually better to avoid having unnecessary controls or buttons to do the same thing, so I think is better to have either the arrows or the drag&drop (whatever is easier or you think would be better).

If we are going to use only the arrows, the should fit in the size of the task-card, so we should remove the padding between them (and maybe make them a bit smaller) I like the arrows you have chosen. I would put them a bit closer to the delete button but overall they look quite good

panvourtsis:
check new icons and functionality
screen shot 2017-10-02 at 15 56 39
screen shot 2017-10-02 at 15 56 35

angrykoala:
The screenshots are looking good.

Sorry, I cannot check the functionality at the moment. I will come back to you when I have a time to try it
Thanks

angrykoala:
@panvourtsis I checked the functionality. The drag behavior is great! and it seems to adapt well to the other interactions (click and the button). A couple of edge cases that we should cover:

* If a task is open (click on the name for edit) the drag behavior is a bit weird. I recommend closing all the tasks when beginning a drag (there is an event/method for this)

* When moving one task, sometimes other tasks also change their order (particularly, moving the bottom task to 2 positions above, swap the tasks in the middle). This seems to be an edge case on how the tasks are reordered.

Everything else seems to be working fine.

panvourtsis:
cool will check fix those and get back at you very soon

angrykoala:
@panvourtsis  Perfect, thanks, I'll close this PR in the meantime

PR #153, opened 2017-10-03
Ordering

panvourtsis:
## Description of the PR
This is a solution to the drag and drop ordering.

### Issues Related
This is targeting the #87 issue

------

**Checklist**
- [x] Your changes have tests

### DIFF
--- a/app/renderer/components/add_task.js
+++ b/app/renderer/components/add_task.js
@@ -3,11 +3,13 @@
 const TaskInput = require('./task_input');
 
 module.exports = {
+    props: ['tasks'],
     data() {
         return {
             title: "",
             command: "",
             path: "",
+            order: 0,
         };
     },
     template: `
@@ -19,7 +21,7 @@ module.exports = {
             </strong>
         </div>
         <div class="collapsible-body container task-card-body">
-            <task-input v-on:save="addTask"></task-input>
+            <task-input v-bind:order="tasks.length + 1" v-on:save="addTask"></task-input>
         </div>
     </li>    
     `,
--- a/app/renderer/components/add_task.js
+++ b/app/renderer/components/add_task.js
@@ -8,8 +8,7 @@ module.exports = {
         return {
             title: "",
             command: "",
-            path: "",
-            order: 0,
+            path: ""
         };
     },
     template: `
--- a/app/renderer/components/add_task.js
+++ b/app/renderer/components/add_task.js
@@ -21,7 +21,7 @@ module.exports = {
             </strong>
         </div>
         <div class="collapsible-body container task-card-body">
-            <task-input v-bind:order="tasks.length + 1" v-on:save="addTask"></task-input>
+            <task-input v-on:save="addTask"></task-input>
         </div>
     </li>    
     `,
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -60,6 +60,7 @@ module.exports = {
     mounted() {
         this.event.on("run", this.run);
         this.event.on("stop", this.stop);
+        this.event.on("collapseTask", this.collapseTask);
     },
     beforeDestroy() {
         this.removeListeners();
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -26,11 +26,15 @@ module.exports = {
     template: `
     <li class="run-card">
         <div class="collapsible-header row unselectable-text">
-            <div class="col s5">
-                <strong class="truncate">{{task.title}}</strong>
+            <div class="col s4">
+                <strong class="truncate">{{task.title}} {{task.order}}</strong>
             </div>
-            <div class="col s3">
+            <div class="col s2">
                 <div class="truncate task-time">{{executionTime}}</div>
+            </div>
+            <div class="col s2">
+                <i v-if="AppStatus.editMode" class="small material-icons" v-on:click="(ev) => { changeItemOrder(ev, 'desc', task) }">arrow_downward</i>
+                <i v-if="AppStatus.editMode" class="small material-icons" v-on:click="(ev) => { changeItemOrder(ev, 'asc', task) }">arrow_upward</i>
             </div>
             <div class="col s3">
                 <a v-if="AppStatus.editMode" class="waves-effect waves-light btn delete-button" v-on:click="onDeleteClick">Delete</a>
@@ -69,6 +73,16 @@ module.exports = {
             ev.stopPropagation();
             this.deleteTask();
 
+        },
+        changeItemOrder(ev, order, task) {
+            ev.stopPropagation();
+            // if (order === "asc") {
+            //   task.order = task.order + 1;
+            // } else {
+            //   task.order = task.order - 1;
+            // }
+            // this.$emit('edit', task);
+            this.$emit('reStructureTasks', order, task);
         },
         deleteTask() {
             this.stop();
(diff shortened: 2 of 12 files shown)


angrykoala:
Hi @panvourtsis Is this the final version of the ordering PR?

panvourtsis:
yes

angrykoala:
I'll do a full review as soon as I can, thanks for your PR

angrykoala:
Ok, after testing it out a bit, seems to be working reasonably good, a few notes on the current behaviour I noted:
* If an edit menu is opened (to edit a task), it should be closed when a reorder starts (ideally it would be automatically be opened afterwards if possible), this is important to allow tasks to be reordered
* A minor glitch where some tasks appear to "jump" sometimes (when there are 3 or more tasks and you try to move one in the middle)

angrykoala:
Please, update the changelog with a new bullet point explaining the changes (a simple one-liner is enough), It doesn't have to be perfectly explained, as I will revise it afterwards, but I think is better that the author describes the feature

app/renderer/components/task_card.js
```diff
@@ -24,9 +24,14 @@ module.exports = {
         "progress-spinner": ProgressSpinner
     },
     template: `
-    <li class="run-card">
+    <li class="run-card task-card">
```
angrykoala: Why is the class task-card added to the li inside a task-card?

app/renderer/components/add_task.js
```diff
@@ -3,11 +3,13 @@
 const TaskInput = require('./task_input');
 
 module.exports = {
+    props: ['tasks'],
     data() {
         return {
             title: "",
             command: "",
             path: "",
+            order: 0,
```
angrykoala: Do we really need the order explicitly inside the task? the order should only affect to suites, and the array of tasks already contains that information

app/renderer/components/task_input.js
```diff
@@ -4,12 +4,13 @@ const Task = require('../task');
...
 
 module.exports = {
-    props: ['task'],
+    props: ['task', 'order'],
     data() {
         return {
             title: "",
             command: "",
             path: "",
+            order: 0,
```
angrykoala: Same as before, I'm not sure we need the explicit order here

app/renderer/components/task_suite.js
```diff
@@ -5,6 +5,7 @@ const EventEmitter = require('events');
 const TaskCard = require('./task_card');
 const AddTask = require('./add_task');
 const AppStatus = require('../app_status');
+const Draggable = require('vuedraggable');
```
angrykoala: Please, move the external dependency (vuedraggable) to the top, just for ordering :)

app/renderer/components/task_suite.js
```diff
@@ -39,6 +41,23 @@ module.exports = {
         AppStatus.events.removeListener("stop-suite", this.onStopSuite);
     },
     methods: {
+        checkMove() {
+            if (AppStatus.editMode) {
```
angrykoala: Why not `return AppStatus.editMode`?

app/renderer/components/task_suite.js
```diff
@@ -39,6 +41,23 @@ module.exports = {
...
+                return true;
+            }
+            return false;
+        },
+        reStructureTasks(event) {
```
angrykoala: restructure is a single word, better naming would be `restructureTasks`

app/renderer/components/task_suite.js
```diff
@@ -71,7 +90,9 @@ module.exports = {
         },
         showAddTab() {
             return AppStatus.editMode && this.suite.length < AppStatus.maxTasksPerSuite;
+        },
+        orderTasks() {
+            return this.suite.tasks.sort((a, b) => a.order - b.order);
```
angrykoala: Instead of using the order property of a task, why not just using an array?

app/renderer/components/task_card.js
panvourtsis: this is here because i set on draggable (task_suite.js line 25 -> :options="{draggable:'.task-card'}") that the class of the option will be that to avoid the "add new" li or future options that we dont need

app/renderer/components/task_suite.js
panvourtsis: not quite sure what do you mean here ...

angrykoala: So, atm you have a set of tasks, and within each task you have a explicit property (order), and you use it to well... order.

However, it doesn't make sense to have that property as part of a task (why would a task care of its own position within different tasks?). In fact, a suite has an array/list of tasks, which, by definition, maintains the order (that was done by design at the beginning). So:

Instead of having:
```js
[{
   name: "second task",
   order: 2
}, {
   name: "first task",
   order: 1
}]
```

And ordering them according to its internal value. We could simply take care in maintaining the array order when writing, reading, moving tasks and rendering:
```js
[{
  name: "first task"
}, {
  name: "second task"
}]
```

We do not need to manually store the order and sort according to it. The array data structure already do that for us, doing this way has some advantages:
* No order property in tasks, no manually ordering each time, the array already contains this information, in general, simpler
* Better for storing/reading data, as we do not need to store the ordering property
* It doesn't make sense to store order in the tasks as it is irrelevant for it, moving a task from one place to another shouldn't change the task itself, but its container (suite)

Basically, I have the impression that solving this issue doesn't require changes to the task.js class, as everything should be part of the suite

I'm not sure if I'm explaining myself, sorry, if you have any doubt I'm happy to make a simple prototype/example to show my point

From what I saw in your code it shouldn't be hard to change it to the way I see it, as most changes you made in the suite are well defined, clean and simple enough to adapt them

panvourtsis:
check my latest commit for the changes we discussed. You were right about the ordering, got a bit confused when we had both arrows and drag but now i think its clean. Let me know what you thing

app/renderer/task_config.js
```diff
@@ -74,7 +74,7 @@ module.exports = {
         return data.map((suite) => {
             let result = new Suite(suite.title);
             result.tasks = suite.tasks.map((task) => {
-                return new Task(task.title, task.path, task.command);
+                return new Task(task.title, task.path, task.command, task.order);
```
angrykoala: I understand that task.order no longer exists. Is this here for some reason or is a typo?

angrykoala:
Everything apart from that extra parameter looks good in the code @panvourtsis

angrykoala:
The code is good, a few details I noticed when running it:

* As I commented before, when starting a drag **all** the tasks should be closed (the edition menu) as it makes harder to actually reorder:

image

* The "drag" behavior is still active when not on edit mode (you cannot actually change the tasks order, but the cards are draggable, this should be trival to solve, adding a class to avoid this when not in edit mode 
img

* The tasks "jump" when reordering sometimes (sorry, I cannot make a screenshot of this) I'm not sure if this is easy or hard to solve, as I feel it is caused by "removing & adding" a task on drag. It may even be because of the library

I'm happy to create separate issues for these problems, as none of them is critical. But if you could solve the ones that you can, I think it would be faster to develop, as you know and remember all the code now

panvourtsis:
create the tasks and i will work on them, assign them to me if you will. Its easier for me also to have it seperate. Thanks

angrykoala:
Ok, I'll create them but GH doesn't allow me to asisgn them to you, please leave a comment to let people know you are going to work in it

Issue #147, opened 2017-09-29
Investigate the use of feature discovery on the edit button


angrykoala:
The materliazecss feature discovery (http://materializecss.com/feature-discovery.html) could be use to point to the edit button if there are no tasks

panvourtsis:
hey, i am working on this. 

Couple of questions. Do you have in mind to popup such message on each refresh or open of the application and when the edit closes with no tasks ?

angrykoala:
Mmm, my idea was to show the popup when the application is **not** in edit mode for a few seconds and the user may not know how that works.

So, maybe we can show it when the following conditions are met:
* The _edit mode_ is **not** active for a few seconds
* The suites are empty
* The edit mode wasn't activated in the session (session = app started)
* The popup wasn't activated in the session

What do you think about those conditions? This task is also a bit of investigation, as I'm not sure if the popup is a good design decision or not.

Also, now it comes to my mind that a simple message in the place where the tasks should be could also be useful (either replacing or along with the feature discovery). Something like "To start adding tasks, press the edit button (<show the edit icon>)

Again I'm not sure if these are good or bad design choices, so I would like to know your opinion

Thanks for taking this issue

panvourtsis:
These conditions were what i was thinking too. But again is a bit annoying to have a popup every time you open the application with no tasks ... Although the popup has a better ui animations i think its more proper to show only a message. Check an image with a demo i made with the popup
<img width="1186" alt="screen shot 2017-10-04 at 12 39 06" src="https://user-images.githubusercontent.com/6433679/31169376-141d3c86-a901-11e7-8793-0392d6bab53f.png">

angrykoala:
Looks good.

Yeah, you are right about the annoying popup. What about adding a plain grey-ish message and having the popup to show only the first time the app is opened? Would that make sense or maybe with the message is enough?.


btw, this idea of one time popup is related to #138

panvourtsis:
yes i've seen it. i agree with this.

panvourtsis:
a question, how do you accomplish first time visit. I though about localStorage value

angrykoala:
We do not need localStorage value, as in electron we can access permanent disk storage.

We are using the module [electron-store](https://github.com/sindresorhus/electron-store) to simplify the configuration storage. You can check the interface to access gaucho config [here](https://github.com/angrykoala/gaucho/blob/master/app/common/app_config.js)

[...]

PR #158, opened 2017-10-05
feature discovery and message on empty tasks

panvourtsis:
## Description of the PR
Implemented a solution for the issue #147 which i introduced a text when tasks are empty to let the user know how to add more. Also on first visit to show a a warning how to navigate 

### Issues Related
#147 

------

**Checklist**
- [ ] Your changes have tests

### DIFF
--- a/app/renderer/components/navbar.js
+++ b/app/renderer/components/navbar.js
@@ -26,7 +26,7 @@ module.exports = {
                     <img class="logo-icon" src="resources/logos/gaucho_logo.png"></img>
                     <a>Gaucho</a>
                 </div>
-                <tap-target v-bind:title="'Add some tasks !'" v-bind:description="'By pressing this button you can add new tasks to your list below.'"></tap-target>
+                <tap-target v-bind:activates="'tap-edit'" v-bind:title="'Add some tasks !'" v-bind:description="'By pressing this button you can add new tasks to your list below.'"></tap-target>
                 <ul class="right navbar-buttons">
                     <li><a id="tap-edit" v-on:click="toggleEdit" v-bind:class="{'edit-button-active': editMode}" class="edit-button"><i class="material-icons unselectable-text">mode_edit</i></a></li>
                     <li><a class="navbar-menu-button" data-activates='navbar-menu' data-gutter="30"><i class="material-icons small unselectable-text">menu</i></a></li>
--- a/app/renderer/components/navbar.js
+++ b/app/renderer/components/navbar.js
@@ -1,3 +1,4 @@
+/* globals $ */
 "use strict";
 
 const Suite = require('../suite');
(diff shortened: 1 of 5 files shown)


angrykoala:
I works fine and looks good. The only minor issue on the functionality is that the discovery displays the edit icon on a different place, when it is closing, for a brief moment both icons are shown.

I'm quite sure this is simply a limitation of the library, and (obviously) is an extremely minor issue, I just comment in case you have an idea to improve it.

I can think of _not_ display the edit button until the discovery is completely close (I guess there is an event for that). But, to be honest, I'm not sure it would look good that way, or if it is worth it, given that it is a brief moment and only once

app/renderer/components/navbar.js
```diff
@@ -22,11 +23,17 @@ module.exports = {
         <nav class="nav-extended">
             <div class="nav-wrapper">
                 <div class="brand-logo main-logo left">
-                <img class="logo-icon" src="resources/logos/gaucho_logo.png"></img>
-                <a>Gaucho</a>
+                    <img class="logo-icon" src="resources/logos/gaucho_logo.png"></img>
+                    <a>Gaucho</a>
+                </div>
+                <div class="tap-target" data-activates="tap-edit">
```
angrykoala: The discovery feature could be a separate component, just to keep navbar small and simple

app/renderer/components/navbar.js
```diff
@@ -1,3 +1,4 @@
+/* globals $ */
```
angrykoala: Please, avoid this, move the method ` $(".tap-target").tapTarget('close');` to the file `materialize.js`, wrap it and use that in this class

If we need to add crappy code because of materialize, at least lets get it all in the same place :)

app/renderer/materialize.js
```diff
@@ -1,14 +1,25 @@
+/* globals $, Materialize */
```
angrykoala: Lovely! I didn't though on this to use jshint in this file

angrykoala:
Another detail about UX, maybe, when the user clicks on the edit button while the popup is active, instead of just closing the popup, the edit should activate (from an user perspective, he is actually clicking the button we are showing him, it doesn't make sense that he has to click it again)

panvourtsis:
about the popup there is no callback function for it and unfortunately on click you can only close it. I dont think though its so much of a problem

app/renderer/materialize.js
```diff
@@ -35,6 +35,9 @@ module.exports = {
             alignment: 'left'
         });
     },
+    toggleTap(selector, status) {
```
angrykoala: Hi, I understand this method is to replace the $ in the other file, but I don't see this being used anywhere in the code

app/renderer/components/tap_target.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+module.exports = {
+    props: ['title', 'description'],
+    template: `
+    <div class="tap-target" data-activates="tap-edit">
```
angrykoala: I understand `data-activates="tap-edit"``references to the tap-edit button outside of the component, shouldn't this be a prop as well so we have a decoupled component?

app/renderer/materialize.js
panvourtsis: you are right i introduced it as a fix to the file but after i reviewed the code from materializecss i've seen that close is not mandatory because when the tap is open the default click behaviour will close it

angrykoala: Oh, nice, then it would be better to remove it

angrykoala:
@panvourtsis about your comment, no, it is not a big deal, just in case it was easy to do, don't worry about that

Issue #94, opened 2017-04-24
Improve unit tests


angrykoala:
Improve unit tests, and reasonably test vue.js components and _front-end_ side

angrykoala:
#105 already covers graphical tests

PR #159, opened 2017-10-05
Adding config for development

### LINKED ISSUE #155: Provide a temporal config for development
Because now config is stored globally, when trying things in dev, you own gaucho config may be affected.

Maybe with a DEV flag or something it would be possible to connect to a different storage than the plain gaucho storage

tdeschryver:
## Description of the PR
Using a dev config when running in dev mode.

### Issues Related
Closes #155.

**Checklist**
- [ ] All the tests and jshint are passing (`npm run test`) - this is giving errors for me, tests are passing when running `mocha` 
- [ ] Your changes have tests - don't know how to set up electron tests to be honest. Once #94 is resolved, I can add tests.

### DIFF
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const Store = require('electron-store');
-
+const { isDevEnv } = require('./utils');
 
 const defaultData = {
     windowSize: [500, 600],
@@ -18,8 +18,9 @@ const key = "ro64wz3l7d";
 
 module.exports = class AppConfig {
     constructor() {
+        const configName = isDevEnv() ? "gaucho_dev_config" : "gaucho_config";
         this.store = new Store({
-            name: "gaucho_config",
+            name: configName,
             defaults: defaultData,
             encryptionKey: key
         });
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const Store = require('electron-store');
-const { isDevEnv } = require('./utils');
+const utils = require('./utils');
 
 const defaultData = {
     windowSize: [500, 600],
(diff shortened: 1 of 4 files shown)


app/common/app_config.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 const Store = require('electron-store');
-
+const { isDevEnv } = require('./utils');
```
angrykoala: Style detail, while this way is generally accepted as a good syntax for requires, I'm using a more basic syntax in the rest of the project (partially for personal preference to be honest), just for consistency. Could you change this for a syntax according to the rest of the code?
Something like this
```js
const utils=require('./utils');
// Rest of the code
utils.isDevEnv()
```

Thanks

main.js
```diff
@@ -4,11 +4,7 @@ const path = require('path');
 
 const MainWindow = require('./app/main/main_window');
 const AppEvents = require('./app/main/app_events');
-
-
-function isDevEnv() {
-    return process.env.NODE_ENV === "dev";
-}
+const { isDevEnv } = require('./app/common/utils');
```
angrykoala: ditto to the first comment

package.json
```diff
@@ -13,21 +13,22 @@
   },
   "devDependencies": {
     "chai": "^4.0.2",
+    "cross-env": "^5.0.5",
```
angrykoala: This is just a question, I understand this is to provide a cross-compatible env, but afaik, envs also work in windows. What is the case we need this?

tdeschryver: When I run `npm run start-dev` without `cross-env` I get the following error:
`'NODE_ENV' is not recognized as an internal or external command,`

app/common/app_config.js
tdeschryver: Yup, no problemo!

package.json
angrykoala: out of curiosity, What OS are you using?

I agree is better to use cross-env to avoid this kind of problems

tdeschryver: Windows 7

Issue #148, opened 2017-09-29
Confirmation on delete


panvourtsis:
Need a popup or something to confirm a deletion on a task

angrykoala:
ditto for suite deletion, and config/clear tasks

A generic popup with confirmation component would be useful to be used on all of these cases.

wildan3105:
Hi, I'd love to work on this. Do you give me a shot?

angrykoala:
Hi @wildan3105, thanks for reaching out. This issue was created by @panvourtsis, so he should decide if he wants to work on this or let you do it.

Meantime, feel free to check all the other issues (or open your own) and work with whatever you prefer

panvourtsis:
hi @wildan3105, please feel free to work on this

panvourtsis:
@wildan3105 are you working on this ?

wildan3105:
Sorry, I had other work to do. You can do by yourself @panvourtsis :disappointed:

panvourtsis:
just checking ! will do after i finish another task. Will let everyone know here

angrykoala:
@wildan3105 If you want to work on this at some point before @panvourtsis starts it, or in any other task/idea whenever you can, feel free to do it :)

panvourtsis:
i made a custom component for modals that we could pass and id title and desciption in order to render. The problem is with materializecss that we need to pre render each of this components in order for it to show them.

Shall we use sweetalert which has a cool integration with vuejs ? https://limonte.github.io/sweetalert2/

angrykoala:
I think there are ways to simplify the problem you say (I basically wrap everything in materialize.js)

I would usually say of fixing stuff instead of adding a new library. But to be honest that library looks great (I didn't know it)

Just a couple of questions:
* Could use that library for every modal (the about menu we currently have)?
* If no, would it make sense to have 2 different styles, one for confirmation using sweetalert and other for other modals?

panvourtsis:
I was suggesting to have every modal on this library 

> this will help us not having "junk" html on the code 

One library would help us fix the current implementation but will also help show messages on the future

angrykoala:
I simply ask about more complex modals (such an about menu or potentially something with a form).

If you think it won't be a problem, go for it, just keep in mind that the about menu should be changed to use this library as well

panvourtsis:
It would easily handle such things. If we agree on this i would change all these modals

angrykoala:
[...]

PR #162, opened 2017-10-06
introduce sweetalert for alerting system

panvourtsis:
## Description of the PR
removing about modal,
introducing new confirm popup on deletion of a task
new cool about modal with sweetalert

### Issues Related
#148 

------

**Checklist**
- [x] Your changes have tests


panvourtsis:
check navbar menu if you think we need to move swal() to some general file or something. Thanks in advance

angrykoala:
Definitely sweetalert makes a better popup than the materialize default. I haven't seen any changes to materialize.js, makes sure all code related to old popups (initialization) is removed.

screenshot from 2017-10-06 15-57-50

I have only a few style things that I thing will improve:
* I don't like all the gap between Version, and by @angrykoala (and, to be honest, I was thinking of taking out that bit) so, for the final version, I think is better to remove `by @angrykoala` completely and reduce a bit the gap
* Version could be better with a different style, something greyish or italics, something a bit less noticeable
* I don't like the yellow box around the link (showing the tab position) please, add a css rule to remove all those highlights from links and buttons of sweetalert. Also, make the pointer the default one (instead of the hand) keep in mind that this is a desktop application, which usually don't have those behaviors.
* I'm not sure Great! (and the :+1: icon) are suitable for that button... I think a `Close` makes more sense

The Confirm delete is perfect, in fact, I would like to add it when deleting a Suite, would you like to make it as part of this or shall I create a different issue?

angrykoala:
Also, there are no changes to the changelog. Please untick that item from the PR checklist, so I'm aware that I should add a line

app/renderer/components/navbar_menu.js
```diff
@@ -1,9 +1,19 @@
+/* globals swal */
```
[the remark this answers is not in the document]
panvourtsis: ok shall i put it on materialize so we can have everything one place then?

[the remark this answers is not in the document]
angrykoala: mmm, not being a materialize thing, I would rather add that to a separate wrapper or change materialize.js name t oavoid confusion

index.html
```diff
@@ -36,6 +36,7 @@
     </script>
     <script src="./node_modules/jquery-ui/build/release.js"></script>
     <script src="./node_modules/materialize-css/dist/js/materialize.min.js"></script>
+    <script src="./node_modules/sweetalert2/dist/sweetalert2.all.min.js"></script>
```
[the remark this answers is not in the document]
panvourtsis: had problems setting it up with require thats why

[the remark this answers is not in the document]
angrykoala: I feel you, materialize was the same

app/renderer/components/navbar_menu.js
```diff
@@ -16,11 +26,24 @@ module.exports = {
         </template>
         <li class="divider"></li>
         <li class="unselectable-text menu-button"><a class="modal-trigger" href="#config-modal">Configuration</a></li>
-        <li class="unselectable-text menu-button"><a class="modal-trigger" href="#about-modal">About</a></li>
+        <li class="unselectable-text menu-button"><a class="modal-trigger" v-on:click="openAbout">About</a></li>
```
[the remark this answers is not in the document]
panvourtsis: in this case though you would have only one component with an <a> link and a method to show the popup ?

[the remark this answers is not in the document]
angrykoala: Yes, either a component or a class, It is better to extract anything that is not strictly of the component.
Navbar should contain the button to show the about, but the about itself and its logic is not part of navbar.

Just a question, I don't know if you already tried, but the internal html of a sweetalert could support components? If so, it would make sense to have the about component, and use it within the sweetalert call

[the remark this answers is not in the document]
panvourtsis: no it doesn't support components. That would be perfect if it was.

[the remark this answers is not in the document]
angrykoala: However, I see you are using v-on in it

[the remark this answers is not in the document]
panvourtsis: maybe i misunderstood. What i am saying is that swal({ html: component }) is not going to work. Is that what you mean also ?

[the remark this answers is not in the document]
angrykoala: No, no, I meant `swal({html: "<about-content></about-content>"})`

This makes a more reasonable thing to have in the method directly

So, in a nutshell, my idea (please, tell me what do you think about this)
1. Wrap swal in a separate class, so we don't have problems with global or compatibility in multiple places
2. Create a component with the content of the about menu and the link logic
3. In navbar, in openMethod, call the wrapped version of swal and simply send an html with the content

Opinions on this?

app/renderer/sweetalert.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+//Methods interface with jQuery
+/* jshint ignore:start */
```
angrykoala: Do we need a full jshint ignore? we can set /* globals swal */ afaik

panvourtsis: dont do a full review yet. Yes of course we can set it.

app/renderer/sweetalert.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = {
+    toggleModal(title, type, children = {}) {
+        return swal(Object.assign({}, {
```
angrykoala: Why do we need an Object.assign(...)?

An idea to improve the wrapper is to add shortcut methods for some common types of modals

And, what are the children?

angrykoala: For what I see in other code, children is more like options, variables or parameters. I think the name children is confusing (I would use it for modals within modals or something like that)

app/renderer/components/navbar_menu.js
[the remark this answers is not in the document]
panvourtsis: Not sure component work on swal 👎  (sadly) so i guess i have to transfer everything inside a component

app/renderer/sweetalert.js
panvourtsis: its a react way of defining children props of a component and that is how i develop/extend components which is actually very common on react. Do you prefer some other naming?

app/renderer/components/task_card.js
```diff
@@ -6,6 +6,7 @@ const TaskStatus = require('../../common/task_status');
 const ProgressSpinner = require('./progress_spinner');
 
 const Material = require('../materialize');
+const SweetAlert = require('../sweetalert');
```
angrykoala: Is SweetAlert a good name? The idea of wrapping the library is that we have a unique interface to alerts that we can change in the future if needed, I'm not sure it makes sense to put the same name than the actual library.

And yes, I'm aware I did the same thing for materialize, which may lead to confusion, but most code in materialize.js is garbage that I [cluttered together](https://en.wikipedia.org/wiki/Diogenes_syndrome) anyway, This alert can be quite useful as a nice wrapper.

app/renderer/components/task_card.js
```diff
@@ -77,8 +78,21 @@ module.exports = {
 
         },
         deleteTask() {
-            this.stop();
-            this.$emit('remove');
+            SweetAlert.toggleModal(
```
angrykoala: The class Alert (or SweetAlert) already scopes the action to a modal (or alert) so I think the method `toggleModal` should be rewritten simply as `toggle`

app/renderer/components/navbar_menu.js
```diff
@@ -1,9 +1,19 @@
 "use strict";
+const {
```
angrykoala: This is not  very important, but in this project, by convention, [I don't use destructing for require statements](https://www.youtube.com/watch?v=6vxHkAQRQUQ)

So this statement should be `const shell=require('electron').shell` to have the same style as the rest of the code

app/renderer/sweetalert.js
angrykoala: You mean the assign?

I'm sorry, I may be a bit lost here, but swal is not a component, is just a method right?

`Object.assign` as you used there is basically creating a copy of the object, this makes sense when the values may be changed (e.g. in a react component) and you don't want them to be changed by the children. This is not the case, as sweetalert is just a method that doesn't change the values.

Making a full copy of the object, effectively, reduces performance and readability, as simply sending title and type as an object should be enough (I may be wrong here, I'm not sure how swal works)


About children naming, basically those are parameters defined by the library like `confirmButtonColor` or `cancelButtonText` right? They are not actually children

app/renderer/sweetalert.js
angrykoala: Oh, sorry, I'm just pointing out the things I saw

app/renderer/sweetalert.js
panvourtsis: Yeah, maybe i explained it wrong. Yes swal is not a component but i was using a component like technique to assign properties from the library. Because i am using different kind of properties on different kind of cases which i dont know and i will not know in the future i created a generic function which takes a title, a type and if you want some children properties to extend your configuration (swal properties) you should pass them as object. This will only take effect if swal has some of these properties of course. It will not affect in any case the result from swal.

I think that object extension is readable enough. It can be more readable in es7 with {...object1, ...object2} but again its simple enough. Also performance wise not sure why you are mentioning this because i dont think that this has some real impact on the application. 

Do you have another idea on how to extend the object with more options for now and for future use? Let me know your opinion in that.

Thanks in advance

angrykoala: Oh sorry, I misread the method Object assign, I though it wasn't being used on children.
You are right, that's the way to go. :+1: 

So, my only comments now are the use of the name "children" and a minor code improvement is that I think you don't need that first argument in assign (you can directly use `Object.assign({title...},children);`

panvourtsis:
had a review in our latest comments and i updated accordingly. Let me know your thoughts

app/renderer/components/about.js
```diff
@@ -12,19 +11,24 @@ module.exports = {
...
-      <div class="modal-footer">
-        <a href="#!" class="modal-action modal-close waves-effect waves-green btn-flat">Close</a>
-      </div>
-    </div>
+      <a v-on:click="openAbout">About</a>
     `,
     methods: {
+        openAbout() {
+            AppAlert.toggle(
+                '<h4>Gaucho</h4>',
```
angrykoala: So, my idea was (and I'm still not sure if it would work), instead of only having a link in the about template, put all this html code as part of the template, and call AppAlert in Navbar.

Something like this:
```js
AppAlert.toggle(html: `<about></about>`);
```
_navbar.js_

This way, the html will be an actual component instead of a plain string in a method, and the about component will be something more than just a link

thought on that? If you don't like it or it is a bit hard to achieve I'm ok with using you current solution

panvourtsis: i though the above and already tried but it doesnt work because it doesn't get translated. That property expects plain html. Do you have any suggestion if this can be skipped somehow?

angrykoala: Ok, I though it may have worked because you were using v-on attributes.

I think everything is correct, I will try and see if I can improve that somehow. I', going to check that everything is working and I'll merge afterwards

angrykoala:
Ok seems to work fine, I'll merge

There are a few possible improvements, I'll create a few issues, as this is working fine as it is

panvourtsis:
will check the issues you will open as well

Issue #175, opened 2017-10-09
Add confirmation on delete suite


angrykoala:
Same confirmation that the one used when deleting tasks should be used when deleting suites

PR #176, opened 2017-10-09
#175 add confirmation on delete suite

arioioio:
## Description of the PR
- Add sweetalert confirmation on suite deletion
- Add dismissal handling on both confirmation dialogs to avoid errors on console

### Issues Related
#175 

------

**Checklist**
- [ ] Your changes have tests

### DIFF
--- a/app/renderer/components/navbar.js
+++ b/app/renderer/components/navbar.js
@@ -5,6 +5,7 @@ const Material = require('../materialize');
 const AppStatus = require('../app_status');
 const NavbarMenu = require('./navbar_menu');
 const TapTarget = require('./tap_target');
+const AppAlert = require('../app_alert');
 
 module.exports = {
     props: ['suites'],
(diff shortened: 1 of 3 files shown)


app/renderer/components/task_card.js
```diff
@@ -92,6 +92,8 @@ module.exports = {
             .then(() => {
                 this.stop();
                 this.$emit('remove');
+            }, () => {
+                return;
```
angrykoala: Do we need the reject callback if we are only returning?

app/renderer/components/navbar.js
```diff
@@ -63,12 +64,27 @@ module.exports = {
...
+                'Are you sure?',
+                'warning',
+                {
+                    text: 'You will not be able to recover this suite after deletion!',
+                    showCancelButton: true,
+                    confirmButtonColor: "#ee6e73",
+                    confirmButtonText: 'Yes, delete it!',
+                    cancelButtonText: 'No, keep it'
+                }
+            )
```
angrykoala: No need to change this, but for next contributions, try to avoid so many almost-empty-lines:
```js
cancelButtonText: 'No, keep it'
}).then(()=>{
```
Is as readable as before, but removes 2 lines :)

app/renderer/components/task_card.js
angrykoala: Fair enough, but in was case is the promise rejected? I haven't checked the library enough yet

angrykoala:
Everything seems to work fine, :+1: 

next time put a comment in the issue before starting working to avoid multiple people working in the same thing

app/renderer/components/task_card.js
arioioio: https://github.com/limonte/sweetalert2#handling-dismissals
another workaround is to use `.catch(swal.noop)`, but then I have to call the swal global variable inside the components, so I just wrote those empty callbacks

angrykoala: No need, your solution seems reasonable, maybe the return may be removed, or the wrapper could be improved so it doesn't throw errors. I was confused on how it worked because rejections are usually reserved for actual errors (not an user response) and I didn't see any error when trying them without catch

Issue #181, opened 2017-10-10
Close "Add Task" from when task saved


angrykoala:
Saving a new task will leave the "Add Task" form open to add a new task. This behavior is a bit inconsistent. Is better to completely close the form, the same way it works when editing a  task

DefecateRainbows:
I'll try this out.

DefecateRainbows:
It's going to take me 8+ hours to get back to this, if someone else wants to do it in that time, that's fine by me

panvourtsis:
i could do it if you want

angrykoala:
Aren't you already in 2 or 3 other issues @panvourtsis ?

panvourtsis:
only task ordering which is very close to this. #156

angrykoala:
If I remember correctly, you also offered to do #161 
I mean, you can work on whatever you want, and it's ok by me if you want to do this issue, but if you are going to take a break from an issue (to work on another thing or simply because of lack of time) send a message so other people are aware and can work on that thing.

Mainly to avoid blockings or people working on the same thing :smile:

panvourtsis:
Yeah we discussed #161 but i didn't know you expected i'll do it. Go confused there!
I will though thanks for clarifying ;)

angrykoala:
Oh, sorry, I misunderstood about #161 .

DefecateRainbows:
Alright, I'm on this.

DefecateRainbows:
Alright, I'm having much more trouble than I should with this. Someone else is going to have to do this D:

ssteuteville:
I'll try tonight.

ssteuteville:
hmmmm.. I have this implemented but I can't get the tests running on Windows 10. Any advice?

`$ npm test

> gaucho@0.3.0 test C:\Users\ssteu\dev\gaucho
> npm run clean && istanbul cover _mocha && npm run jshint


> gaucho@0.3.0 clean C:\Users\ssteu\dev\gaucho
> rm -rf builds config.json tasks.json coverage

No coverage information was collected, exit without writing coverage information
C:\Users\ssteu\dev\gaucho\node_modules\.bin\_mocha.CMD:1
(function (exports, require, module, __filename, __dirname) { @IF EXIST "%~dp0\node.exe" (
                                                              ^
SyntaxError: Invalid or unexpected token
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
[...]

PR #185, opened 2017-10-11
issue-181 add task should close

ssteuteville:
## Description of the PR
We collapse the header for add_task when a new task is added. Created a util method to avoid duplicating code. Let me know if you want that or anything else changed. Also, I wasn't able to get the tests running on windows (I left a comment on the issue) and since this is a DOM change I didn't see any examples of how you'd want this tested.

Happy to add unit test for the DOM manipulation if you can provide me an example.. Or we could move the util method into a new module that would be shared between components to help with code coverage, but that feels like we're just hiding lack of coverage. Let me know what you want.

### Issues Related
#181 

------

**Checklist**
- [?] Your changes have tests

### DIFF
--- a/app/common/utils.js
+++ b/app/common/utils.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const Material = require('../renderer/materialize');
+
 module.exports = {
     generateTimeString(time) {
         const seconds = time % 60;
@@ -26,5 +28,12 @@ module.exports = {
     },
     isDevEnv() {
         return process.env.NODE_ENV === "dev";
+    },
+    collapseHeader($el) {
+        const elements = $el.getElementsByClassName('collapsible-header');
+        if (elements[0]) {
+            elements[0].classList.remove("active");
+            Material.updateCollapsible();
+        }
     },
 };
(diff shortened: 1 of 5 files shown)


ssteuteville:
ok got the tests running locally with this command:
`istanbul cover node_modules/mocha/bin/_mocha -- -R spec` 

But I also noticed you have CI to run the tests as well.

app/common/utils.js
```diff
@@ -27,4 +29,11 @@ module.exports = {
     isDevEnv() {
         return process.env.NODE_ENV === "dev";
     },
+    collapseHeader($el) {
```
angrykoala: I like the idea of creating this function, but wouldn't it make more sense having this in materialize (or maybe elsewhere)?

keep in mind that utils.js (and everything under common) is intended to be used (or at least, to not depend) on either renderer or main proceses, while `collapseHeader` is exclusively a renderer process

angrykoala:
I checked that this PR solves #181 correctly, thanks for contributing @ssteuteville 
I commented about a possible improvement in the code, please, check the comment before I merge to discuss possible changes :+1:

ssteuteville:
I was thinking the same thing after I submitted PR. Code is updated!

Cheers

PR #187, opened 2017-10-11
Minor cleanup and refactoring

### LINKED ISSUE #186: Minor code improvements
There are minor improvements to perform in the code overall (or things that should be checked):

- [x] app/renderer/suite.js line 64: Extract the string inside existTaskName to a separate variable with a proper name
- [x] main.js line 16: Instead of parsing a string, check if the url module can be used
- [x]  app/renderer/suite.js line 61: `let` used instead of `const`. Check if there is a reason for this

> Lines may be aproximate

kushalpandya:
### This PR has following changes:

**Changes proposed in #186 are as follows:**

- ~Instead of extracting string concatenation to a variable at [`app/renderer/suite.js#L63`](https://github.com/angrykoala/gaucho/blob/master/app/renderer/suite.js#L63), it converts concatenation to ES6 template string~ This is already updated on `dev` and is appropriate choice here instead of extracting to variable, so I've just removed unnecessary whitespace.
- Uses `url` module to construct `htmlUrl` used in [`main.js#L20`](https://github.com/angrykoala/gaucho/blob/master/main.js#L20)
- Usage of `let` instead of `const` at [`app/renderer/suite.js#L61`](https://github.com/angrykoala/gaucho/blob/master/app/renderer/suite.js#L61) is appropriate as we're modifying `index` instead `while` loop.

**Additional improvements:**

- Adds constant config object in `app/main/main_window.js` to replace direct string usages of `windowSize`, `devToolsSize` & `maximized` at multiple places.
  - Reason is that hard-coding strings in multiple places is generally prone to typos as updating such string would require to update multiple places.
- Converts `let` to `const` in `app/renderer/suite.js` in multiple places where const is appropriate instead of let.

### Issues Related

Fixes #186

### DIFF
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -3,6 +3,12 @@
 const Store = require('electron-store');
 const utils = require('./utils');
 
+const FIELDS = {
+    WINDOW_SIZE: "windowSize",
+    DEVTOOLS_SIZE: "devToolsSize",
+    MAXIMIZED: "maximized",
+};
+
 const defaultUserConfig = {
     windowSize: [500, 600],
     devToolsSize: 300,
@@ -75,6 +81,7 @@ class TasksConfig extends AppConfig {
 }
 
 module.exports = {
+    FIELDS,
     User: UserConfig,
     Tasks: TasksConfig
 };
--- a/app/main/main_window.js
+++ b/app/main/main_window.js
@@ -27,7 +27,7 @@ module.exports = class MainWindow {
 
     initWindow(devWindow) {
         let win;
-        const windowSize = this.userConfig.get("windowSize");
+        const windowSize = this.userConfig.get(AppConfig.FIELDS.WINDOW_SIZE);
 
         let winConfig = {
             width: windowSize[0],
@@ -40,7 +40,7 @@ module.exports = class MainWindow {
             frame: true
         };
         if (devWindow) {
-            winConfig.width += this.userConfig.get("devToolsSize");
+            winConfig.width += this.userConfig.get(AppConfig.FIELDS.DEVTOOLS_SIZE);
         }
 
         win = new BrowserWindow(winConfig);
@@ -50,24 +50,24 @@ module.exports = class MainWindow {
 
         if (devWindow) win.webContents.openDevTools();
 
-        if (this.userConfig.get("maximized") === true) {
+        if (this.userConfig.get(AppConfig.FIELDS.MAXIMIZED) === true) {
             win.maximize();
         }
 
         win.on('resize', () => {
-            if (this.userConfig.get("maximized") !== true) {
+            if (this.userConfig.get(AppConfig.FIELDS.MAXIMIZED) !== true) {
                 let size = win.getSize();
-                if (devWindow) size[0] -= this.userConfig.get("devToolsSize");
-                this.userConfig.set("windowSize", size);
+                if (devWindow) size[0] -= this.userConfig.get(AppConfig.FIELDS.DEVTOOLS_SIZE);
+                this.userConfig.set(AppConfig.FIELDS.WINDOW_SIZE, size);
             }
         });
 
         win.on('maximize', () => {
-            this.userConfig.set("maximized", true);
+            this.userConfig.set(AppConfig.FIELDS.MAXIMIZED, true);
         });
 
         win.on('unmaximize', () => {
-            this.userConfig.set("maximized", false);
+            this.userConfig.set(AppConfig.FIELDS.MAXIMIZED, false);
         });
 
         let first = true;
--- a/app/renderer/suite.js
+++ b/app/renderer/suite.js
@@ -11,7 +11,7 @@ class Suite {
     }
 
     addTask(task) {
-        let title = this.getValidName(task.title);
+        const title = this.getValidName(task.title);
         task.title = title;
         this.tasks.push(task);
     }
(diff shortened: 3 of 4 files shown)


angrykoala:
Please, make sure you were updating the last version of gaucho:dev, there are conflicts (and, in fact, it shouldn't be a .jshintrc anymore)

kushalpandya:
@angrykoala Yes just realized it :sweat_smile: I'll rebase and update PR. :slightly_smiling_face:

kushalpandya:
@angrykoala Branch updated, also removed commit related to `.jshintrc`, all yours now. :wink:

app/renderer/suite.js
```diff
@@ -43,7 +43,7 @@ class Suite {
     }
 
     runAll() {
-        for (const task of this.tasks) {
+        for (let task of this.tasks) {
```
angrykoala: Why let here?

app/main/main_window.js
```diff
@@ -1,5 +1,11 @@
 "use strict";
 
+const CONFIG = {
```
angrykoala: If we are going to use a set of constants for config values, this should be in common/AppConfig so it can be used by all the places of the app

app/renderer/suite.js
```diff
@@ -58,15 +58,15 @@ class Suite {
     getValidName(name) {
         let index = 2;
         if (!this.existTaskName(name)) return name;
-        while (this.existTaskName(`${name  } (${  index  })`)) {
+        while (this.existTaskName(`${name} (${index})`)) {
```
angrykoala: `${name} (${index})` should be assigned to a constant with a name

main.js
```diff
@@ -13,7 +14,11 @@ function initApp() {
     function createWindow() {
         if (win === null) {
             const iconPath = path.join(__dirname, 'resources', 'icon.png');
-            const htmlUrl = `file://${__dirname}/index.html`;
+            const htmlUrl = url.format({
+                pathname: path.join(__dirname, 'index.html'),
+                protocol: 'file:',
+                slashes: true
```
angrykoala: Is slashes needed?

app/renderer/suite.js
kushalpandya: It is part of `for...of` loop where on every iteration, `task` variable would be updated from one of the value of `this.tasks` list. Although we're not updating `task` inside the loop, it is generally considered good practice to use `let` within `for`.

app/renderer/suite.js
kushalpandya: At the end of every iteration, we're increasing value of `index` by one, so if we make `${name} (${index})` part of a const, we'd have to change structure of this look such that assigned variable is also updated at the end of loop. I can do the change, but it would be slightly more for refactoring :thinking: what do you suggest?

angrykoala: Well, not necessarily a const but a let.

mmm, you got a point there a do...while could solve this, but indeed, probably the result is worse than the problem. Lets leave it this way then

app/renderer/suite.js
angrykoala: mmm, I always though that using const was better than let, even in these cases, as we are creating a new task variable each iteration.

I don't really mind either way, but I'm not sure on what the good practice is here

main.js
kushalpandya: Not necessary, but https://electron.atom.io/docs/tutorial/quick-start/ suggests this as default way to contruct URL, although looking at [`url.format`](https://nodejs.org/api/url.html#url_url_format_urlobject) documentation, I don't think we need `//` appended to the end result, I can remove it. :+1:

app/main/main_window.js
kushalpandya: You mean using constants as `AppConfig.WINDOW_SIZE`, etc?

main.js
angrykoala: please, do it, everything else looks good

app/main/main_window.js
angrykoala: that or AppConfig.configValues.WINDOW_SIZE

whatever you think best

app/renderer/suite.js
kushalpandya: Generally, if variable will not be modified during the scope of it, `const` is considered ideal, other `let`. And `let` usually goes with `for` as most of the times it gets modified within, however, in this particular case, we can use `const` as well since `task` is never modified within the scope, but having it `let` would be good in case we modify task later. :slightly_smiling_face:

app/main/main_window.js
kushalpandya: Yup makes sense, having `AppConfig.CONFIG.WINDOW_SIZE` would make sense as we'd not have to export anything else apart from AppConfig. I'll update this PR. :+1:

kushalpandya:
@angrykoala I've updated this PR based on our discussion. :slightly_smiling_face:

app/common/app_config.js
```diff
@@ -75,6 +81,7 @@ class TasksConfig extends AppConfig {
 }
 
 module.exports = {
+    CONFIG,
```
angrykoala: Maybe `FIELDS` (or `fields`) instead of `CONFIG` (mainly to avoid AppConfig.CONFIG which is a bit redundant

app/renderer/suite.js
angrykoala: Well, my point here is that if we modify the task later, a const will fail, so we simply change to let. While we don't change the variable we could just put const.

app/common/app_config.js
kushalpandya: Yup, makes sense, I'll update to `FIELDS` instead.

kushalpandya:
@angrykoala I've updated this PR to use `AppConfig.FIELDS` instead of `AppConfig.CONFIG`. Also, I've left usage of `let` in `for...of` as it is as if in future we plan to modify `task` within loop, we won't be touch first line. :slightly_smiling_face:

PR #192, opened 2017-10-14
Use `material-design-icons-iconfont` package from npm in place of local stylesheet

### LINKED ISSUE #182: Remove materialicons from repository
Currently, under resources, in the repo there are all the material icons.
This shuldn't be on the repository, and they should be installed/downloaded somehow in the install or building process

kushalpandya:
## This PR does following:

- Add [`material-design-icons-iconfont`](https://www.npmjs.com/package/material-design-icons-iconfont) package from NPM that provide iconfont-only version of Material Icons
- Removes local version of Material Icons from `resource/material_icons` directory.

## Reasoning:

Having local copy of Material Icons is hard to keep up-to-date especially when NPM is already available to fix it.

Google already has [`material-design-icons`](https://www.npmjs.com/package/material-design-icons) package available on NPM which includes iconfonts but it also includes all the vector and raster versions of each icon, which makes up the download size of over **40 MBs**, and given that it is run-time dependency, it isn't practical to include 40 MB of extra unused assets in packaged app.

While Google recommends to use Google Fonts API to load Material Icons, it isn't useful when user wants to use Gaucho offline. Hence, [`material-design-icons-iconfont`](https://www.npmjs.com/package/material-design-icons-iconfont) is ideal replacement which is just iconfonts-only version of the package which Google offers.

Fixes #182


kushalpandya:
@angrykoala Please review. :slightly_smiling_face:

angrykoala:
+2 −2,412, lovely.

Thanks, I didn't know of that library, everything seems to work fine

Issue #21, opened 2016-11-28
Import task option


angrykoala:
Add an option to import a gtask.json file.

Importing the file will (for now) override the current suites and tasks, so an alert ( #162 ) should be used to let the user know and confirm if that's what he wants.

> Remember to stop **all** executing tasks when doing this, as they will be removed, this must be done **after** the user confirmed the alert

angrykoala:
currently using tasks.json, investigate changing the filename to gtask.json

angrykoala:
Changed issue name

angrykoala:
Because export is already implemented (#23 #150 ) this issue is important for the following release (we need an import for the export). Do you want to take this issue @grg121 ?

grg121:
yep, it's mine

ssteuteville:
mb put wrong issue number on my pull request.. I'm not working on this.

PR #195, opened 2017-10-14
Import tasks

grg121:
## Description of the PR
I've done beautiful things.

### Issues Related

#21 import tasks option


app.js
```diff
@@ -30,12 +30,16 @@ const app = new Vue({ // eslint-disable-line no-unused-vars
...
         AppStatus: AppStatus
     },
-    components: components,
+  components: components,
     mounted() {
         Material.init();
         TaskConfig.loadTasks();
         this.suites = TaskConfig.suites;
         this.loaded = true;
+        const updateApp = function(){
```
angrykoala: Why not using an anonymous function?
```
AppStatus.events.on("update-suites",()=>{
    this.suites=TaskConfig.suites;
});
```

angrykoala: Note the this.suites instead of app

app/renderer/#app_status.js#
```diff
@@ -0,0 +1,23 @@
+"use strict";
```
angrykoala: Why is there a new file with # in the name?

Please, make sure you remove you editor auto-generated files or add them to .gitignore

app/renderer/components/#task_card.js#
```diff
@@ -0,0 +1,147 @@
+"use strict";
```
angrykoala: same as before

app/renderer/components/.#task_card.js
```diff
@@ -0,0 +1 @@
+grg121@arch-greyhound.2129:1507985031
```
angrykoala: another weird file

app/renderer/components/config_menu.js
```diff
@@ -49,8 +50,25 @@ module.exports = {
...
-    methods: {
+   methods: {
+        importTasks() {
+            dialog.showOpenDialog({
+                filters: [{
+                    name: 'json',
+                    extensions: ['json']
+                }]
+            }, (fileNames) =>{
+                if (fileNames === undefined) return;
```
angrykoala: if (!filenames) is a better check, as it checks for null as well

Probably you should check that fileName[0] exists as well

app/renderer/components/config_menu.js
```diff
@@ -49,8 +50,25 @@ module.exports = {
...
+              let fileName = fileNames[0];
```
angrykoala: couldn't this be a const?

app/renderer/task_config.js
```diff
@@ -18,6 +19,12 @@ module.exports = {
         }
         this.suites = this.parseData(suites);
     },
+  loadTasksFrom(data){
```
angrykoala: This is not indented at the same level as rest of the code

Use tab=4 spaces

resources/styles.css
```diff
@@ -307,3 +307,7 @@ html, body {
     width: 100%;
     margin-top: 10px;
 }
+
+em{
```
angrykoala: should this affect to every em tag? maybe it should have it's own class or use css nesting (`.config-menu em{` or something like that)

app/renderer/components/config_menu.js
```diff
@@ -49,8 +50,25 @@ module.exports = {
...
-    methods: {
+   methods: {
+        importTasks() {
+            dialog.showOpenDialog({
+                filters: [{
+                    name: 'json',
+                    extensions: ['json']
+                }]
+            }, (fileNames) =>{
+                if (!filename || !FileNames[0]) return;
```
angrykoala: empty return is not very nice, maybe the opposite condition and wrapping everything else inside the if?
```
if(filenames && fileNames[0]){
...

```

Also, a couple of things:
1. Use a beautifier or something to avoid incoherent tabs, this if statement is not properly aligned
2. Should the first element be `fileNames`? I mean, filename is not yet defined 
3. filename is just one word, it should not be formatted as fileName

app/renderer/components/config_menu.js
```diff
@@ -63,6 +81,7 @@ module.exports = {
                         console.warn(err);
                     });
                 }
+
```
angrykoala: empty line is empty

app/renderer/components/navbar.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
     mounted() {
         Material.checkFirstTimeTap(".tap-target");
     },
-    methods: {
+  methods: {
```
angrykoala: wrong alignment

angrykoala: [js-beautify](https://www.npmjs.com/package/js-beautify) is a good package for javascript automatic formatting, try that if your editor is not formatting js properly

app/renderer/components/config_menu.js
```diff
@@ -58,16 +58,18 @@ module.exports = {
...
-              if (!fileNames[0]) return ;
-              const fileName = fileNames[0];
-              const confirmationAlert = new DeleteConfirmationAlert("You will not be able to recover this task after deletion!");
-              confirmationAlert.toggle().then(() => {
-                TaskConfig.clearTasks();
-                fs.readFile(fileName, 'utf-8', (err, data) =>{
-                  TaskConfig.loadTasksFrom(data);
-                });
-            }, () => {});
+              if (fileNames){
```
angrykoala: `if (filenames && filenames[0])` better

Also, filename instead of fileName please

Issue #26, opened 2016-12-07
Context menu


angrykoala:
right-click button should display a context menu with options

angrykoala:
This package could be useful: https://github.com/sindresorhus/electron-context-menu

ssteuteville:
I'm interested. What should the options be to start?

angrykoala:
Well, we could start simple and start integrating new things afterwards, so, for example:
* **About**: Will open the about menu, this option appears at the bottom everywhere you click
* **Toggle Edit**: Will get you in edit mode (or out of it), this option appear everywhere
* **Run/Stop**: Will run/stop the task, this option will **only** appear when clicking on top of an app (if possible, make Run/Stop option change accordingly) and while **not** in edit mode

What do you think for a start? If you see these options are too many to begin with, I'm happy to separate some of them in a different issue afterwards.

If you have any ideas, feel free to comment and implement them as well

Remember that these options should be relatively easy to modify or add new ones

ssteuteville:
Ok cool, I think I could get something like that put together in a few days. Assign, please.

angrykoala:
I can't assign issues to people who are not direct colaborators, anyway, your comment should be enough, consider yourself assigned.

If you need some help, or have something working event without all the features, just comment or PR

ssteuteville:
Having some trouble with this. I tried the library you suggested but I haven't figured out how to get it to have a different context based on what you right clicked.. I've also tried vue-context-click, but it's not handling clicks properly for some reason. Starting to think this one might be too involved for me if I can't get one of these libraries working.

angrykoala:
Hi @ssteuteville , Indeed, I tried the electron-context-menu library and it doesn't seem to support custom context menus depending on class.

If there is no built-in library for this, a possible solution is either to use bare electron or creating an html-css context menu. Both these options require considerable more work and investigation, so if you prefer to check other issues that are simpler it's ok

thanks for your help with this, I'll check another options and come back to this issue

angrykoala:
Here there is a working example of an html-based context menu, may be of help:
http://qnimate.com/customizing-right-click-in-html-page/

ssteuteville:
I'll take another stab at this. I'm using this as a learning exercise for vue and electron and also I want the hacktoberfest shirt -- so my skills in those areas are sub par.

If I can't get this done in a few hours I'll prob switch to a less involved task. I'll try again tonight!

angrykoala:
[...]

Issue #196, opened 2017-10-14
Sortablejs warning


angrykoala:
![screenshot from 2017-10-14 19-05-25](https://user-images.githubusercontent.com/5960567/31577628-b699ebf8-b112-11e7-98ed-08b091faa7d5.png)

Sortable library used for tasks reordering throws a warning, probably not important, but worth investigating related to #156

angrykoala:
may be related to #202

rkomiyama:
It doesn't seem to be related to #202, since I made a quick fix for `Sortable.js` and that warning is still there.

I've put up a [PR](https://github.com/RubaXa/Sortable/pull/1202) in the `Sortable` repo for the fix. If they aren't okay with those changes, we could use our own fork of their project and make the appropriate changes.

angrykoala:
@rkomiyama Thanks for digging into this. I don't think it is worth it making out own fork (this is just a minor issue), but it would be awesome if they accept your PR so we could get rid of this warning.

Please keep this issue updated with any news about this

angrykoala:
This seems to be a problem of Sortable library, closed for now

Issue #178, opened 2017-10-09
Update mocha to version 4


angrykoala:
Mocha (the testing framework) is currently on 3.2.0 in the package.json, updating to version 4 will make the tests freeze, investigate compatibility issues and check if it can be updated

Issue #156, opened 2017-10-05
Task ordering improvements


angrykoala:
The ordering defined and implemented in #87 and #153 can be improve, fixing the following minor issues

- [x] When starting a drag **all** the tasks should be closed (the edition menu) as it makes harder to actually reorder:
- [x] The "drag" behavior is still active when not on edit mode (you cannot actually change the tasks order, but the cards are draggable, this should be trival to solve, adding a class to avoid this when not in edit mode 

- [ ] The tasks "jump" when reordering sometimes (sorry, I cannot make a screenshot of this) I'm not sure if this is easy or hard to solve, as I feel it is caused by "removing & adding" a task on drag. It may even be because of the library

* ~~When reordering a running task, the log will remain in the previous position~~ Check #213 
![screenshot from 2017-10-08 22-38-15](https://user-images.githubusercontent.com/5960567/31320719-7b097b96-ac79-11e7-8544-281d7507d088.png)
As you can see, the only running task (redis) is showing the log in the previous position. This problem is confirmed to happen when deleting a task above the running task as well, so it'll probably require some extra implementation, as it is not exclusive of task reordering

- [x] Reordering should only work when dragging in the task card (name and drag icon), not when dragging in the form. This makes impossible to copy the text in the form nor in the run button

panvourtsis:
i am on it !

angrykoala:
Hi @panvourtsis because this task is quite big, if you need some help with it just write a comment here.
Keep in mind that I'm happy to accept a PR with any of the fixes, no need to solve all of them in one go

PR #197, opened 2017-10-15
Context menu

ssteuteville:
## Description of the PR
Context Menus

### Issues Related
#26 

------

**Checklist**
- [] Your changes have tests


angrykoala:
The current behavior of the context menu in the normal view feels good. good job

Indeed there are some issues with the draggable behavior, i'm not surprise it's messing with the menu. You can check some of the problems I detected with it in #156 and #196 I don't know if it will be of your help.

I don't really understand why its messing, but probably because of a event capture by the library, my guess is that the draggable is attached to the whole element, and not only the card or the icon (known issue described in #156) you can try disabling the dragabble item or setting it only to the "drag" icon to see if the behavior can be fixed.

Another idea, I can see draggable is used in task-suite component. You could capture every event to that component (in the parent component to the draggable). You can check there the type of event and prevent before affecting the draggable. 
It is a bit messy, but it would mean sending a custom event to the task, so it opens the context-menu, instead of relying in the click event.
So:
```
click->suite(captures click)
     if right click
          preventsDefault()
          suite->send_custom_event->task->opens_menu
     else
          do nothing (event will be captured by draggable)
```
This is just an idea, I hope it helps

Overall I think you are going in the good direction, as the context menu in the run view works quite well.
I know this is not the final version, but a few comments that may help you:
* The context menu should be improved, I recommend using (or copying, if it is not possible tu use it directly) the same style as the main menu (however, keep the size the same, the current menu is too big for a context menu)
* It may be useful creating a vue component for the context menu, it will help in development and reusablity.
* Avoid adding methods and stuff like that to AppStatus, that class should only be used for global-scoped variables

ssteuteville:
Ok thanks for input I'll try to have something deliverable tonight.

ssteuteville:
still on this just been busy.

angrykoala:
No worries, there is no rush :)

ssteuteville:
@angrykoala I think this should be in a decent state now. I saw your comment about the AppStatus, however, since the method's primary functionality is mutate AppStatus, I couldn't think of a better place to put it. I'm open to suggestions but I think it makes sense to keep the business logic with the model.

I needed to track the currentContextMenu in appStatus because I needed access to it globally in the application to close it when a new one opened.

angrykoala:
Keeping the business logic with the model is specifically a bad design imo
For now (while it may not be ideal) I would put that logic in the component itself, AppStatus should just be data.
I'll do a full review when I can, but please, change that for now
Good job!

angrykoala:
btw, you were able to fix the problem with the edit mode?

ssteuteville:
Moved the open context menu logic. No, I played with the click events for too long and was not able to fix the edit mode thing :(

angrykoala:
I don't see big issues with the code, I'll confirm it is working and merge when I can
thanks :+1:

angrykoala:
I works mostly fine, we can leave the edit problem for another future, however, there are some usage problems that I see:
* When selecting run/stop in a task the menu doesn't hide
* Menu won't open out of the tasks or navbar (white space or bottom bar)
* Menu should have a smaller font and button size
* Menu displays out of the window if clicking in the border

Do you think you may be able to fix these? (or most of these)

ssteuteville:
I'll work on these changes over the weekend.

ssteuteville:
@angrykoala I've fixed all of the issues you brought up, got rid of my code that turned draggable off and on since somebody cleaned up the draggable implementation, and made significant (imo) improvements to the code by removing duplicate code and added default behavior to the context-menu component.

Hope this can get merged :D

(when you merge can you point me to some easy issues for my hacktoberfest goal, I still need 2 more)

angrykoala:
Hi @ssteuteville , there are a few issues with the tag `good-first-issue` that should be relatively easy and quick to do. Both were already "assigned" but none of them did a PR, so feel free to ask in the comments if they are still working (if you are interested)

The issue #178 should be quick as well

Fell free to start working on them whenever you want, you don't have to wait for this PR to be accepted

angrykoala:
Hi @ssteuteville 
Reviewing your code I see you created an events singleton in GauchoActions. I think it is a good idea, but currently (we are still in the middle of a code refactoring) there is already a events singleton in `AppStatus.events`. Is there a reason for you not to use that singleton?

As I see, you made several changes. I will probably need some time to merge those changes, as some of them are improvements to the code we want to maintain, while other may collide with other changes I'm making right now. I'll try to merge this as soon as I can

app/renderer/components/task_card.js
```diff
@@ -66,7 +69,9 @@ module.exports = {
     },
     methods: {
         toggleRun(ev) {
-            ev.stopPropagation();
+            if (ev.target.classList.contains('btn')) {  // without this check the context menu doesn't close
```
angrykoala: I don't fully understand this comment, and why this check

ssteuteville:
I considered the same thing for the context menu but the solutions I came up with seemed more complicated than they were worth. I think the way it is now is more clear what is going on and when a context menu will have what values. If you have suggestions I'm open to hearing them out!

app/renderer/components/task_card.js
ssteuteville: Since I'm using `toggleRun` for the context menu click handler and also for the `btn` click handler it was stopping propagation for the context menu as well which kept it from closing. The context menu closes on click because the event propagates and this code runs:

https://github.com/angrykoala/gaucho/pull/197/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR37

angrykoala:
mmm, I can only think on using some global storage (like app status) and/or events to communicate a click and the context menu.

The main problems I see here is:
1. More instances of the component mean more possibilities of collision and worse performance
2. I feel we will end with duplicated code in the future with this approach

All in all, you may be right and any other solution may end up being more complex, I'll have to tinker a bit and see if I get an improvement. I'll try to merge this in a separate branch anyway and try to improve it afterwards

index.html
```diff
@@ -14,25 +14,32 @@
...
     <section id="app">
+        <context-menu ref="menu"></context-menu>
         <config-menu></config-menu>
-        <header id="main-header" class="margin-bottom" style="height:100px">
+        <header id="main-header" class="margin-bottom" style="height:100px" v-on:contextmenu.prevent="$refs.menu.openContextMenu">
             <navbar v-bind:suites="suites"></navbar>
         </header>
 
-        <main v-bind:class="{ 'full-size': !AppStatus.config.bottomBar }">
+        <main v-bind:class="{ 'full-size': !AppStatus.config.bottomBar }" v-on:contextmenu.prevent="$refs.menu.openContextMenu">
```
angrykoala: Why is this event needed?

index.html
```diff
@@ -14,25 +14,32 @@
...
-        <footer v-if="AppStatus.config.bottomBar">
+        <footer v-if="AppStatus.config.bottomBar" v-on:contextmenu.prevent="$refs.menu.openContextMenu">
             <bottom-bar></bottom-bar>
         </footer>
     </section>
 
     <script>
         window.$ = window.jQuery = require('./node_modules/jquery/dist/jquery.min.js');
         window.Hammer = require('./node_modules/materialize-css/js/hammer.min.js');
+        window.document.body.addEventListener('click', function (ev) {
```
angrykoala: do we need this hardcoded in the html?

index.html
ssteuteville: That is the right click event that triggers opening the menu.

I had to put it on all 3 of these elements because $refs was undefined on the section and body tags.

index.html
ssteuteville: No I think that could probably go into app.js

angrykoala: Ok, please, put it in app.js if possible and I think that's all, thanks

angrykoala:
Hi @ssteuteville , I see this PR only requires a quick fix related to some code that should be moved, are you able to do it?

app/renderer/api/gaucho_actions.js
```diff
@@ -2,10 +2,24 @@
 
 const AppStatus=require('../app_status');
 const TasksHandler=require('../tasks_handler');
+const EventEmitter = require('events');
+const GauchoEvents = new EventEmitter(); //EventEmitter singleton
```
angrykoala: Please, use AppStatus.events instead

ssteuteville:
@angrykoala I've got really busy over the past week. I should have some time this weekend to get this branch up to date and make the changes though.

angrykoala:
Hey @ssteuteville , I'm planning on releasing Gaucho 0.4 soon, Initially I wanted to add this to the release, but I'm not sure if this is going to be ready for then (it is a complex feature that requires a lot of testing). What do you think (how much time do you think we need for this to be ready)?

ssteuteville:
Hey sorry life got kinda crazy for a little bit! Let me get this branch up to date, make the changes you asked for, and give this some testing.

I think the functionality is scoped well enough right now that the testing effort shouldn't be too much.

I'll update you in a few hours when this is done.

ssteuteville:
Ok done. All code review changes are made.

I tested all of the context menu functionality.

I also did regression testing on the following: dragging tasks around, starting/stopping tasks, using the top bar options, switching suites, and adding a new task.

Hope this helps with your release.

angrykoala:
Thanks, I'll check this as soon as I can. Looks great

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 0.4.0 / 2017-10-02
 ==================
 
+  * Added context menus to the application
```
angrykoala: From now on, don't change the changelog (no problem here, I'll change it myself) as there are conflicts among contributors

app/renderer/components/context_menu.js
```diff
@@ -0,0 +1,42 @@
...
+        openContextMenu(event) {
+            if (AppStatus.currentContextMenu !== null) {
+                AppStatus.currentContextMenu.$el.style.display = 'none';
+            }
+            AppStatus.currentContextMenu = this;
+            let menuElement = this.$el;
+            menuElement.style.display = "block";
+            let bounds = menuElement.getBoundingClientRect();
+            // Math.min here to avoid the menu going off the side of the screen
+            menuElement.style.left = `${Math.min(window.innerWidth - bounds.width - 3, event.clientX)}px`;
```
angrykoala: Could this be in a separate method? something like _calculateStyles

app/renderer/components/context_menu.js
```diff
@@ -0,0 +1,42 @@
...
+            </li>
+        </ul>
+  `,
+    methods: {
+        openContextMenu(event) {
+            if (AppStatus.currentContextMenu !== null) {
+                AppStatus.currentContextMenu.$el.style.display = 'none';
+            }
+            AppStatus.currentContextMenu = this;
+            let menuElement = this.$el;
```
angrykoala: These should be `const`

app/renderer/components/context_menu.js
```diff
@@ -0,0 +1,42 @@
...
+            event.stopPropagation();
+        }
+    },
+    computed: {
+        _menuItems() {
+            if (this.menuItems) { // if menu items are set use them
+                return this.menuItems;
+            }
+            // otherwise use default
+            return [{name: 'About', click: GauchoActions.toggleAbout },
```
angrykoala: The default should be a separate variable (e.g `const defaultActions`) the comments here are redundant

resources/styles.css
```diff
@@ -379,3 +379,20 @@ html, body {
...
+.context-menu
+{
+    min-width: 100px;
+    display: none;
+    opacity: 1;
+    position: fixed;
+}
+
+.context-menu li.menu-item {
+    min-height: 0px;
```
angrykoala: 0 shouldn't have `px` (just 0 is better)

resources/styles.css
```diff
@@ -379,3 +379,20 @@ html, body {
...
+}
+
+.context-menu li.menu-item a {
```
angrykoala: Do we really need 3 nested styles? coudln't we simply assign a class to the element we want to style?

angrykoala:
I found 2 usage problems:
* Secondary click over the menu trigger the default menu to reopen (e.g. clicking twice when over a card). Secondary click should be disabled when hovering over the context menu
* Pressing `esc` should close the context menu

Overall, it is looking better, probably the few changes I pointed out will be enough to have a first version of the context menu

angrykoala:
Closed due to inactivity

Issue #157, opened 2017-10-05
Move tasks between suites


angrykoala:
Task can be reordered within a suite (#87). However, it is not possible to move tasks across suites. Investigate how this could be accomplished

sergeir82:
I am on it

PR #199, opened 2017-10-15
Moving tasks between suits.

sergeir82:
### Move tasks between suites #157

Move task to suite header an interface switch on this suite.


app/renderer/components/add_task.js
```diff
@@ -13,18 +13,23 @@ module.exports = {
...
-            <strong class="unselectable-text">
-                <span class="small material-icons">add</span>
-                Add New Task
-            </strong>
-        </div>
-        <div class="collapsible-body container task-card-body">
-            <task-input v-on:save="addTask"></task-input>
-        </div>
-    </li>    
+    <ul class="run-card collapsible no-margin" data-collapsible="accordion" >
```
angrykoala: This "addtask" was intended to be part of the same accordion, so when opening the AddTask button, it will close the others.

I would prefer to maintain AddTasks as li, instead of a whole new `ul`. I'm not completely sure why you are using a new ul, but if it is needed, you have to close every opened task in that suite

app/renderer/components/task_suite.js
```diff
@@ -8,7 +8,7 @@ const AddTask = require('./add_task');
 const AppStatus = require('../app_status');
 
 module.exports = {
-    props: ['suite', 'index'],
+    props: ['suites', 'index'],
```
angrykoala: a question, why do you need all the suites here? I assume because it is needed for the library. Can you think of a solution where the task drag is out of the task suite component so a suite doesn't need all the other suites?

I'm ok with your solution, I want to know for possible improvements in the code

app/renderer/components/task_suite.js
```diff
@@ -25,18 +25,23 @@ module.exports = {
...
-            <add-task v-bind:tasks="suite.tasks" v-on:add="addTask" v-if="showAddTab"></add-task>
             </draggable>
+            <add-task v-bind:tasks="suites[index].tasks" v-on:add="addTask" v-if="showAddTab"></add-task>
         </div>
     `,
     mounted() {
         AppStatus.events.on("run-suite", this.onRunSuite);
         AppStatus.events.on("stop-suite", this.onStopSuite);
-        AppStatus.totalTasks += this.suite.length;
+        AppStatus.totalTasks += this.suites[this.index].length;
```
angrykoala: I see you are using suites[this.index] a lot, probably is better to set a variable with the current suite?

I would use the `data()` method, returning the current suite

app/renderer/components/add_task.js
sergeir82: When the add button was in the tasks ul tag and it was no task in this list, I could not drag any tasks to it. You can see this if checkout to ddf686a1b4c189bc8c0451c3201684ad25d15d5e. I think around and fix it that way.

angrykoala: Any idea why that happens when using the same ul?

angrykoala:
Hi @sergeir82 , after testing your changes again to merge the PR, I found a problem,I cannot add task easily as there is a problem with the form when the text is empty, I need you to fix this before I merge your changes

angrykoala:
Are your changes ready for review?

angrykoala:
There is an unwanted change in the style of Add New Task:

screenshot from 2017-10-22 23-09-52
_Your Version_
screenshot from 2017-10-22 23-10-10
_Current Version_

Is there a reason for this change or it was introduced unintentionally?

app/renderer/components/add_task.js
```diff
@@ -13,22 +13,27 @@ module.exports = {
...
-            <strong class="unselectable-text">
-                <span class="small material-icons">add</span>
-                Add New Task
-            </strong>
-        </div>
-        <div class="collapsible-body container task-card-body">
-            <task-input v-on:save="addTask"></task-input>
-        </div>
-    </li>    
+        <li class="no-draggable"  v-on:pointerdown="preventDragStart($event)"
```
angrykoala: The problem related to the style can be because you removed the run-card class from the element

angrykoala: Do we really need all this changes to add_task? it wasn't possible to drag it before either

angrykoala:
Hi @sergeir82  Everything seems to be working fine, just 3 things:
1- There is a conflict, I can fix it afterwards if you want, so you don't have to worry about it
2- I feel that some changes to AddTasks are not necessary, vuedraggable already takes care of elements that cannot be dragged, and I'm not sure if there is another reason for those changes
3- There is a small style change in the add tasks button, probably because of a missing class

Thanks for your contribution!

app/renderer/components/add_task.js
sergeir82: This only fix problem with form when the text is empty. Vuedraggable don't care about forms in filtered elements.

sergeir82:
1- Tried merge by myself. Hope nothing lost.
2- Found that Sortable plugin has option preventOnFilter that default true which cause problem https://github.com/RubaXa/Sortable#options. Set false and removed mousedown handlers from AddTasks.
3- Class was added.

angrykoala:
Thanks @sergeir82, I'll check this as soon as I can

angrykoala:
@sergeir82 Everything looks good, thanks. One question, What is `group: 'tasks'` option for?

Issue #188, opened 2017-10-11
Add confirmation alert for 'clear tasks' action


angrykoala:
Use the same alert wrapper as suite and task delete

ccalamos:
Do you mean when deleting a task?

angrykoala:
No, that already exists, when entering conguration, you have the action of "clear all tasks". That should have a similar alert/modal than the current modal displayed when deleting a task or suite (to check this, make sure you are on dev branch and updated)

ccalamos:
I can replicate the same modal as requested, however there are some errors within clearing tasks itself that should be addressed as well such as the ticker not decreasing to zero.

angrykoala:
What ticker? You mean the tasks list? That should be fixed, make sure you are running latest `dev` branch, if the problem persist, please, open a new issue detailing the problem and how to replicate

Also, replicating the modal should be trivial, as our current wrapper already does that for you, so basically is running that wrapper before executing the action of the button. Do you want to do it @ccalamos ?

techpool:
Hi @angrykoala, I would like to take up this task and proceed with it. Can you point out the location where I should start looking into?

angrykoala:
Sure @techpool , the code to display alerts is in `app/renderer/app_alerts` you shouldn't even need to modify it, but in case you need to, it should be easy to extend and export a new class from AppAlert, the `DeleteConfirmationAlert` however should be enough for this issue.

In `app/renderer/components/config_menu.js` you have the component of the config menu where you can find the `clear task` button action.

Remember to checkout to `dev` branch before working on this

angrykoala:
@ccalamos Maybe the problem you reported is the same as #190 ? If so, it should be already fixed

grg121:
HI, @ccalamos @techpool Are you still working on this?

ccalamos:
@grg121 Sorry I was away from my computer this weekend, @angrykoala yes it is the same as issue #190 and I can do this by tonight however I do not want step on @techpool's feet if he is currently working on this. If not I would like to claim it and submit a PR by tonight.

techpool:
Same here @grg121. @ccalamos you can go ahead and create the PR. I'm not sure if I'll be able to make it my tonight. Thanks.

angrykoala:
Ok, so @ccalamos will work on this, thanks @techpool and @grg121

PR #201, opened 2017-10-15
#188 Add confirmation alert for 'clear tasks' action

ccalamos:
Related to issue #188, quick easy solution to add confirmation for clearing tasks.

### DIFF
--- a/app/renderer/components/config_menu.js
+++ b/app/renderer/components/config_menu.js
@@ -10,7 +10,7 @@ const AppStatus = require('../app_status');
 const SwitchForm = require('./switch_form');
 const TasksHandler = require('../tasks_handler');
 const TaskImporter = require('../../common/task_importer');
-
+const DeleteConfirmationAlert = require('../app_alerts').DeleteConfirmationAlert;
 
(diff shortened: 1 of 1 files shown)


angrykoala:
@ccalamos One question, it was not defined in the issue, but it would be cool if the config menu close  after confirming the "clear tasks" action (and not when cancelling)

Do you think it is easy for you to do it as part of this PR or it would be better to open a separate issue for this?

ccalamos:
I will look into it 👍

angrykoala:
Everything looks fine, but closing the menu when clicking clear task is weird, please, remove the `modal-action modal-close` from the button class to avoid this behaviour

ccalamos:
So to confirm you want the confirmation modal to appear on top of the config menu and then if you say yes to bring you to main screen and if you say no then show config modal?

angrykoala:
Yes, that's the ideal behavior if you can do it with relative ease
Thanks!

ccalamos:
I will work on having it not close out but I say we finish this PR and open a new issue and I will continue to work on it.

Issue #203, opened 2017-10-15
Config menu should close when confirming clear task


angrykoala:
Only when confirming, not before or when canceling
Related to #188

ccalamos:
Quick question on this... Your eslint does not include jQuery definitions, I can change it in this PR to do so, however the easiest way to perform this will require me to use the '$' which will throw a un-def error because you have not selected jQuery

angrykoala:
You can use /* globals $ */ or something similar I think

PR #204, opened 2017-10-16
Issue #203 Config menu should close when confirming clear task solved.

ccalamos:
Only thing to note that ESLint was changed to approve of jQuery Variables such as '$'

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,6 @@
     "mocha": true,
     "node": true,
     "es6": true,
-    "jquery": true
   },
   "plugins": ["vue"],
   "globals": {},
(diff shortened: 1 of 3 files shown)


angrykoala:
Thanks, Your solution seems to be working fine. However a small comment about $.

Because JQuery is only used as a _hack_ for the materialize library. Currently every use of it is wrapped in the file `app/renderer/materialize.js` to avoid high coupling between gaucho and Jquery. In this file you will see how $ is defined as a global in a comment (to avoid eslint errors)

So, please, remove the changes in eslint and create a method in materializecss (e.g. `closeModal(modalSelector)` and use it from config_menu. This way all the hacks for materialize library are in one place :)

ccalamos:
Sounds great!  Just updated it if you want to take a look.

Issue #55, opened 2017-03-22
Improve log behavior 


angrykoala:
When opening a log, it is scrolled up at the beggining, change this behaviour so it starts at the bottom

angrykoala:
A bug introduced at some point broke any functionality related to the log being scrolled down

ssteuteville:
Poked at this a little and got a fix. Going to submit a PR since there is no activity here yet.

PR #207, opened 2017-10-18
log should auto scroll

ssteuteville:
## Description of the PR

The problem was that when autoScroll ran from the print function, it didn't have any affect if the log wasn't open. So, I added an event listener for when the body opens so that we can scroll for the user on that event as well. During development I found that the next dom update ($nextTick) was firing to soon for us to change the scroll on the div (collapsible.js wasn't done running). So, I swapped out to next javascript event loop (setTimeout 0) and it started working.

Not sure if setTimeout is really the right way to go here or if there is just something about vue that I'm missing. Anyways, this fixes the bug I think.

### Issues Related
#55 

------

**Checklist**
- [ ] Your changes have tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 0.4.0 / ####-##-##
 ==================
-
+  * Auto scroll working on logs
   * Tasks can now be reordered within a suite while in edit mode
   * Added an initial message, noting the edit button to new users
   * Tasks with no path defined will now be executed on user home directory
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -26,7 +26,7 @@ module.exports = {
     },
     template: `
     <li class="run-card task-card">
-        <div class="collapsible-header row unselectable-text">
+        <div class="collapsible-header row unselectable-text" v-on:click="autoScroll">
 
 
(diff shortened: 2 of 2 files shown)


angrykoala:
mmm, the solution is a bit hacky, but resonable.

Could you check the collapsible api? (http://materializecss.com/collapsible.html) There is an event for onOpen collapsible. May it be useful for this? Just an idea in case it can be improved

angrykoala:
Your solution has a couple of situations that don't work as I would expect as an user:
1- If I open a log and the task keep sending output, that scroll doesn't go along  with the outpu
2- If I open a log, scroll up, close the log, and open again, it is scrolled down.

So, the scroll should be down in the following situations:
1- When opening a task, by default, is down, unless the user already opened the task and scrolled up
2- If there is new output, and I'm at the bottom, it should scroll down along with it.

This conditions, reset when the task is re-run

angrykoala:
Yes, sorry about that, the thing is that the other features were already working some time ago whene I wrote that issue, but some of the broke, so now the UX is a bit weird.
Thanks!

angrykoala:
This PR doesn't perform the required behavior, I'll close this (it'll be easier to start from scratch to avoid the conflicts)

Issue #213, opened 2017-10-22
When deleting a task or reordering, the log is not updated


angrykoala:
When reordering tasks or deleting a task, the log position is not updated and it will still be displayed in the old position. Related to #156

panvourtsis:
Hi @angrykoala, what do you mean with this ?

angrykoala:
Let's imagine 2 tasks in a suite (`echo 'hello'` as first task and `ls` as second) If you execute both, you will see the logs of each one by clicking in the card.
If you now swap them and open the log, the `hello` log will still be in the top card

panvourtsis:
ok i will work on this. Will check and let you know

PR #214, opened 2017-10-24
Log update

panvourtsis:
## Description of the PR
Fixing the problem reading the correct log on reordering running tasks. It was reading the item's output but the item might not had changed but the task would. Now the output although is not on the task model will be stored in it temporarily. This is good for future use in other parts if needed as well.

### Issues Related
#213
------

**Checklist**
- [ ] Your changes have tests


angrykoala:
Hi @panvourtsis 
Your solution works well, and I like the idea of storing the output in the task. Could you please add the output (and the logic of making it empty on run) as part of the Task model itself? This should also simplify the task component
Thanks!

panvourtsis:
something like this you mean?

angrykoala:
Yes, but the print action of concatenating outputs (the `this.task.output=\n....`) could also be part of the task model

However, because it may be out of the scope of this ticket, I'm happy to accept this PR and create a new follow-up issue for improvements

panvourtsis:
no its fine leave it here and i will work on it

panvourtsis:
fyi i removed the config (config.outputMaxSize) which was not in use to not use unused libs and configs as well

angrykoala:
Hi @panvourtsis , I detected a problem with you current version
* When a task that logs something while it is running, the logs won't appear until the task finish logging (try with a command like `ping www.google.com`, you'll see it won't log anything until you stop it, I think this is because you added the log handling (`this.output = `\n${out.join("\n")}`.trim();`) at the end of execution, not for each stdout. You should add it just before calling `autoscroll` in the Task class.

* Also, the config.outputMaxSize was used to determine the maximum log size ` this.output = this.output.slice(-config.outputMaxSize).trim();` , this is important, as a huge log will heavily affect performance.

Thanks for your help

panvourtsis:
having a small issue with the above although i fixed everything because i am requiring app_status.js on the task.js i am getting an error on when running test from scripts.

Do you have any clue ?

```
/Users/panagiotisvourtsis/Documents/workspace/gaucho/node_modules/electron-store/index.js:8
                const defaultCwd = (electron.app || electron.remote.app).getPath('userData');
                                                                   ^

TypeError: Cannot read property 'app' of undefined
    at ElectronStore (/Users/panagiotisvourtsis/Documents/workspace/gaucho/node_modules/electron-store/index.js:8:54)
```

angrykoala:
What version of  electron are you running? They made a few breaking changes. Please, update your branch with the latest changes of `dev` and do a clean install (remove node_modules and execute `npm install`) I already changed the version in package.json

If the problem persists, please tell me

panvourtsis:
done that but still ...

angrykoala:
I tried executing the tests in your branch and I wasn't able to reproduce your problem
Please tell me:
1. Commands you used to encounter this error
2. Version installed of electron-store and mocha
3. your node and npm versions (`node -v` and `npm -v`)

angrykoala:
Also, I'm having a conflict here, so I assume you didn't merge dev into your branch

panvourtsis:
electron 1.7.9
mocha 4.0.1
node 6.0.0
npm 4.1.2

angrykoala:
Ok, I got the problem, seems to be an issue with electron-storage package, probably a broken version I'll investigate that

thanks for noticing

angrykoala:
Ok, now I understood, sorry, I'm a bit slow.
I don't like the idea of reuqiring AppStatus form a common file. So, at least for now, I'm happy to use a variable defined in task.js for the size
Please, remove `outputMaxSize` from the config and use a variable hardcoded for now, this should make everything easier

panvourtsis:
tests are failing though, have you seen this?

panvourtsis:
Tasks
    ✓ Valid Tasks Status
    1) Correct Task Execution
    2) "after each" hook

  3) "before each" hook

angrykoala:
I think is the same problem, you have to remove the AppStatus.js dependency on task.js

For now, just hardcode the outputMax variable in task.js and remove it from the config

panvourtsis:
sorry forgot to tell that i did and i get this

angrykoala:
I still see that dependency in the PR. Are you sure you removed it from `common/task.js`?

If so, please show me the full log

angrykoala:
Hi @panvourtsis , this is working great, and no problems with the code, but tests are failing. Probably because this PR changes the structure of the Task class, which may affect some tests. Shouldn't be hard to fix, please, fix them before I merge

angrykoala:
I fixed the tests for you, It was just a minor change

panvourtsis:
thanks for that

Issue #218, opened 2017-10-25
Make timers optional


angrykoala:
Add an option to config so the timers are displayed or not

If the timers are disabled, check if the performance can be improved by removing unnecessary events.
If the timers are enabled, the timers **must** have the correct values

angrykoala:
![screenshot from 2017-10-25 15-52-29](https://user-images.githubusercontent.com/5960567/32002322-86b53392-b99c-11e7-940b-68b43b66898a.png)

angrykoala:
![screenshot from 2017-10-25 15-52-54](https://user-images.githubusercontent.com/5960567/32002341-95ee3ab6-b99c-11e7-8071-22a4d95aef73.png)
_Example of timers_

theTechie:
on it

PR #220, opened 2017-10-25
Make timers optional

theTechie:
Enables the feature mentioned in #218.

We should now be able to configure the visibility of the timer from the config menu.


angrykoala:
Please, make the merge to the `dev` branch

Issue #219, opened 2017-10-25
Adding keyboard shortcuts


angrykoala:
Would it make sense to have keyboard shortcuts in gaucho? Either pre-defined shortcuts to move and execute tasks or configurable keys

davidsonsns:
I believe key shortcuts can be a very interesting feature. Makes use much more convenient and fast.
If you agree, I'll start this feature.

You can attribute to me :smile:

angrykoala:
Feel free to start with it. I recommend doing something simple, so we all can test it and see if it is working fine

Keep in mind that this may not be trivial to implement (depends on how electron and vue handle those kind of things)

davidsonsns:
Great, I'll do it.

PR #222, opened 2017-10-26
Adding keyboard shortcuts

davidsonsns:
### #219 
Adding keyboard shortcuts to validate its implementation

------

**Checklist**
- [ ] Your changes have tests


app/renderer/components/navbar.js
```diff
@@ -97,6 +101,24 @@ module.exports = {
                 default:
                     this.AppStatus.events.emit(selection);
             }
+        },
+        addShortcuts() {
```
angrykoala: 2 improvements here:
1. Could you please wrap all Mousetrap usage in a separate file and use that interface here instead please?
2. `$('#navBarMenu').click()` Is a jquery event, for now I'm wrapping all these events in the file `materialize.js` to avoid high coupling and make future refactorings easier. Could you please wrap this event as well?

app/renderer/components/navbar.js
```diff
@@ -1,5 +1,8 @@
+/* globals $ */
```
angrykoala: Remove this after fixing the other comment

angrykoala:
Hi @davidsonsns 
Sorry for taking so long to answer. The keyboard shortcuts behavior is great, thanks.
I just added a comment for a minor improvement in the code. Everything else is great
Thanks

davidsonsns:
I will make the changes.

I will try to enable the arrows to move between the options, especially in the menu, usability will be much better.

angrykoala:
Please, confirm this PR is final (I saw you didn't added the arrows, I'm happy to add that in a separate issue)

app/renderer/api/materialize.js
```diff
@@ -53,5 +53,11 @@ module.exports = {
     },
     closeModals() {
         $('#config-modal').modal('close');
+    },
+    getLengthNavBarTabs() {
```
angrykoala: you don't really need to do this. Navbar is the size of AppStatus.suites, I think is better and cleaner using that, instead of relying on jQuery

app/renderer/api/materialize.js
```diff
@@ -53,5 +53,11 @@ module.exports = {
...
+        return $('#navbar-tabs .tab-button').length
+    },
+    toggleNavBarMenu() {
+        $('#navBarMenu').click()
```
angrykoala: I think materialize already provides a [method](http://materializecss.com/dropdown.html) to open/close dropdowns, It is cleaner using that instead of making a click

app/renderer/api/shortcuts.js
```diff
@@ -0,0 +1,45 @@
...
+    init() {
+        this.changeTabForward();
+        this.changeTabBack();
+        this.toggleEdit();
+        this.toggleNavBarMenu();
+    },
+    changeTabForward() {
+        Mousetrap.bind('ctrl+tab', () => {
+            const totalTabs = Material.getLengthNavBarTabs() - 1
+            const newIndex = AppStatus.activeSuite === totalTabs
```
angrykoala: This code is a bit hard to read (using a ternary operator when the condition is not simple makes it hard to read, maybe an if is more readable?

If not, please, write a single line for the operator

app/renderer/api/shortcuts.js
```diff
@@ -0,0 +1,45 @@
...
+                : AppStatus.activeSuite + 1
+
+            Material.selectTab("#navbar-tabs", `tab${newIndex}`);
+            GauchoActions.toggleActiveSuite(newIndex);
+        });
+    },
+    changeTabBack() {
+        Mousetrap.bind('ctrl+shift+tab', () => {
+            const totalTabs = Material.getLengthNavBarTabs()
+            const newIndex = AppStatus.activeSuite === 0
```
angrykoala: same as before

app/renderer/api/shortcuts.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+    changeTabForward() {
+        Mousetrap.bind('ctrl+tab', () => {
+            const totalTabs = Material.getLengthNavBarTabs() - 1
+            const newIndex = AppStatus.activeSuite === totalTabs
+                ? 0
+                : AppStatus.activeSuite + 1
+
+            Material.selectTab("#navbar-tabs", `tab${newIndex}`);
+            GauchoActions.toggleActiveSuite(newIndex);
```
angrykoala: I see a lot of repeated code in this method and changeTabForward

maybe part of the logic could be extracted into a third common method?

app/renderer/components/config_menu.js
```diff
@@ -44,6 +44,14 @@ module.exports = {
                     <a class="waves-effect waves-light btn" v-on:click="importTasks">Import Tasks</a>
                     <label><em class="warning-text">ALERT! this will override your previous tasks</em></label>
                 </div>
+
+                <div class="shortcuts-learn">
```
angrykoala: This could be a separate component (no need to fix it for now, as it is small enough to be part of config), just as an idea. What do you think?

app/renderer/api/materialize.js
davidsonsns: I can not find suites in appstatus, would you confirm? so I inserted with jquery the length.

angrykoala: Sorry, it is in TaskHandler

davidsonsns: I really enjoyed your tips. as I am new to the Vue, the manipulation of the state does not dominate very well :smile:

app/renderer/api/materialize.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         $('ul.tabs').tabs();
     },
     updateDropdown() {
-        $('.navbar-menu-button').dropdown({
+        $('#navBarMenu').dropdown({
```
angrykoala: html classes and ids should be separated by hyphen by convention:
e.g. `navbar-menu`

app/renderer/api/materialize.js
```diff
@@ -53,5 +53,9 @@ module.exports = {
     },
     closeModals() {
         $('#config-modal').modal('close');
+    },
+    toggleNavBarMenu() {
+        const elNavBarMenu = $('#navBarMenu')
```
angrykoala: to be honest, after having a second look to the docs and the resulting code, I think it was better as $(...).click() the toggle. Sorry for making you change this again

app/renderer/components/config_menu.js
```diff
@@ -43,6 +46,8 @@ module.exports = {
                     <a class="waves-effect waves-light btn" v-on:click="importTasks">Import Tasks</a>
                     <label><em class="warning-text">ALERT! this will override your previous tasks</em></label>
                 </div>
+
+                <shortcuts-learn />
```
angrykoala: Is autoclosing tag supported in vue?
Make sure it is (I though you need to use open-closing tags)

app/renderer/components/navbar.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
                 <tap-target v-bind:activates="'tap-edit'" v-bind:title="'Add some tasks !'" v-bind:description="'By pressing this button you can add new tasks to your list below.'"></tap-target>
                 <ul class="right navbar-buttons">
                     <li><a id="tap-edit" v-on:click="toggleEdit" v-bind:class="{'edit-button-active': editMode}" class="edit-button"><i class="material-icons unselectable-text">mode_edit</i></a></li>
-                    <li><a class="navbar-menu-button" data-activates='navbar-menu' data-gutter="30"><i class="material-icons small unselectable-text">menu</i></a></li>
+                    <li><a id="navBarMenu" class="navbar-menu-button" data-activates='navbar-menu' data-gutter="30"><i class="material-icons small unselectable-text">menu</i></a></li>
```
angrykoala: I'm thinkusing the id `navBarMenu` is misleading. Keep in mind that this is not the actual menu, but the button.

If you need an id, we could turn the class `navbar-menu-button` into an id (making sure it is not being used anywhere else) and modifying the related css

app/renderer/components/config_menu.js
davidsonsns: really autoclosing is bad
[Self closing tag](https://vuejs.org/v2/style-guide/#Self-closing-components-strongly-recommended)

angrykoala:
Everything looks fine. Thank you

Issue #211, opened 2017-10-22
Export and import tasks should close the configuration menu


angrykoala:
**Only** after the process of exporting and importing was successful (i.e. no problems and the user confirmed the action. The config menu should automatically close, the same way it is closed in 'clear tasks' action

yasirazgar:
looking into it

PR #231, opened 2017-10-27
#211 -> close config menu when import/export

grg121:
Solve #211 
Now config menu is closed only if the action of import/export is confirmed.
Also I've added the same behaviour for the 'clear taks' buttom.

:)

### DIFF
--- a/app/renderer/components/config_menu.js
+++ b/app/renderer/components/config_menu.js
@@ -13,19 +13,19 @@ const DeleteConfirmationAlert = require('../api/app_alerts').DeleteConfirmationA
 const Materialize = require('../api/materialize');
 
 module.exports = {
-  data() {
-    return {
-      config: {
-        bottomBar: AppStatus.config.bottomBar,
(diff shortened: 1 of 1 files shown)


app/renderer/components/config_menu.js
```diff
@@ -67,6 +67,7 @@ module.exports = {
                         cancelButtonText: "No, cancel import"
                     });
                     confirmationAlert.toggle().then(() => {
+                        ('#config-modal').modal('close');
```
angrykoala: This is a materialize method, in this project (to avoid [ñapa code](https://calidadysoftware.wordpress.com/2015/07/10/naas-napa-as-a-service/)) I'm wrapping all materialize crap inside the file `materialize.js`

In fact, there is already a wrapped method for this (closeModals). Please use that method instead

angrykoala:
I just created a new branch beta-0.4
Please, merge in that branch instead of dev

Issue #225, opened 2017-10-26
Make PKGBUILD for Arch Linux (git version)


aledomu:
I was thinking about making an Arch's build script to automatically package gaucho directly from this repo. I could make another PKGBUILD for stable releases when 0.4 is released.

This script would go in this repository, so if you agree with writing the script for the git version, a PR will follow soon.

angrykoala:
Hi @aledomu 
What kind of packages are generated by PKGBUILD?

Keep in mind that for packaging we will be using [electron-builder](https://github.com/electron-userland/electron-builder) to create the packages. Unless strictly necessary I would like to avoid having separate pipelines for build.

afaik, Arch uses `pacman` as package manager, which is among the supported packages from electron builder:
```
snap, deb, rpm, freebsd, pacman, p5p, apk, 7z, zip, tar.xz, tar.lz, tar.gz, tar.bz2, dir
```

If you want to add support for it, instead of making a separate PKGBUILD, try modifying the `package.json` file in `dev` (https://github.com/angrykoala/gaucho/blob/dev/package.json) to add pacman (or any other) to the building process according to the electron-builder docs.

Please, check if the resulted package after running `npm run dist` is working in Arch. 
If it is not working, we may consider adding a separate pipeline with a script to provide support for that platform.

Thanks

aledomu:
Ok, I'll tweak the `package.json` file so it builds Arch packages. We'll discuss later all the stuff about the PKGBUILDs.

angrykoala:
ok

angrykoala:
#233 merged, I assume this is not enough to close this ticket, what else do we have to do?

aledomu:
I initially thought it was a good idea to put the PKGBUILD in this repo, but seeing how other people accomplish the same task I think it's better to just send it to AUR and the one who wrote that is who maintains it. So it's safe to close this as I see nothing more to do here.

angrykoala:
Adding wontfix tag as no changes were made as part of this issue

PR #233, opened 2017-10-28
Added pacman target and enabled 32-bit builds

aledomu:
AppImage was untested, so it was deleted. Also changed the portable target in Windows to zip since it's less prone to error. <https://github.com/angrykoala/gaucho/issues/225>

### Issues Related

------

**Checklist**
- [ ] Your changes have tests


package.json
```diff
@@ -40,22 +40,25 @@
...
+    "dist-win": "npm run clean && electron-builder -w --x64 --ia32",
+    "dist-mac": "npm run clean && electron-builder -m"
   },
   "build": {
     "appId": "angrykoala.gaucho",
     "linux": {
       "target": [
         "deb",
         "rpm",
-        "AppImage"
```
angrykoala: Why did you remove AppImage build?

package.json
```diff
@@ -40,22 +40,25 @@
     "build-electron-all": "npm run clean && npm run build-electron -- --platform=win32 && npm run build-electron -- --platform=linux && npm run build-electron -- --platform=darwin",
     "clean": "rm -rf builds config.json tasks.json coverage dist",
     "lint": "eslint *.js app test",
-    "dist": "npm run clean && electron-builder -wlm"
+    "dist": "npm run clean && electron-builder -wlm --x64 --ia32",
+    "dist-linux": "npm run clean && electron-builder -l --x64 --ia32",
+    "dist-win": "npm run clean && electron-builder -w --x64 --ia32",
+    "dist-mac": "npm run clean && electron-builder -m"
```
angrykoala: --x64 and --ia32 are not needed here?

package.json
```diff
@@ -40,22 +40,25 @@
...
-        "AppImage"
+        "pacman"
       ],
       "category": "Utility"
     },
     "win": {
       "target": [
         "nsis",
-        "portable"
+        "zip"
```
angrykoala: Why did you remove portable?

package.json
aledomu: It's just to build for both x64 and x86 architectures. By default it will only build for the architecture you are running at compile-time (usually x64).

package.json
aledomu: It needed some metadata that wasn't present and kept failing.

package.json
aledomu: Packaging in zip deploys a portable version as well, it just doesn't make a first-run executable, so you launch it directly and you don't depend on any external tool in the packaging process. Although seeing it now, I think it's a good idea to switch the order between nsis and zip, so if nsis fails (happens sometimes in Linux), at least the zip target is built.

package.json
angrykoala: Is that error part of the errors in https://github.com/angrykoala/gaucho/wiki/Common-Issues ?

I didn't have that error, I would like to maintain AppImage support unless this is not fixable

package.json
angrykoala: After trying, I can see that the generated versions are:
* win-unpacked: standalone executable in the folder, this is generated always, for every system in the build
* Gaucho setup (nsis), installer for windows
* `zip`: same as unpacked but in a zip

The portable version is a single file that can be executed, while indeed, you can use the unpacked version (or the zip directly), I would like to maintain the "portable" version as a true standalone single-portable-file

package.json
angrykoala: Yeah, I know, my question is why not using it in `dist-mac`?

aledomu: macOS just works on 64 bits.

package.json
aledomu: Nope, none of those. I could get it back and let you test it (if it worked for you previously, it should still work).

package.json
aledomu: Ok, will change.

package.json
angrykoala: Could you show me the error? (I already tested your version with AppImage and it seemed to work, I'll try again)

angrykoala:
Please @aledomu , check the issue #238 for discussion about 32 bits

package.json
aledomu: It shows me this:
`./dist/__appimage-x64 should be packaged as /home/AJ/hacktoberfest/gaucho/dist/gaucho-0.4.0-x86_64.AppImage`

package.json
```diff
@@ -35,24 +35,27 @@
     "test": "istanbul cover _mocha -- --exit && npm run lint",
     "start": "electron .",
     "start-dev": "cross-env NODE_ENV=dev npm start",
-    "clean": "rm -rf builds config.json tasks.json coverage dist",
+    "clean": "rm -rf config.json tasks.json coverage dist",
     "lint": "eslint *.js app test",
-    "dist": "npm run clean && electron-builder -wlm"
+    "dist": "electron-builder",
+    "dist-all": "npm run clean && electron-builder --x64 --ia32"
```
angrykoala: A few concerns:
* What does `electron-builder` (without arguments) do?
* Why not adding npm run clean to dist?
* Don't we need the -wlm option in `dist-all`?

aledomu: 1. That `dist` script is there to build for all platform specified in `package.json` in the architecture currently running the script.
2. The `dist` script is thought to be more "manual" so to speak, it is meant to be used passing arguments so we don't have to make a script for each platform and/or target. Adding `npm run clean` would remove previous builds even if you are building to a platform not already built.
3. Not at all, it would package for all platforms and targets configured in `package.json`. Also this way you can specify to build to just one platform and/or target in both currently supported architectures, although it's discouraged because this script does remove all previous builds.

If you don't agree with the rationale, feel free to ask more questions or ask for specific fixes.

angrykoala: 1. Ok, I don't understand why, by default, `dist` only generates current OS build, while `dist-all` generates all by default without other arguments

2. I agree with passing arguments, but I'm not sure how those arguments are supposed to be passed, could you try and document this in the readme?

aledomu: I have tested it and I must have done something wrong, so dist-all has the same issue, my bad. So we do need `-wlm` in `dist-all`.
Passing arguments works this way, for example, `npm run dist -- --ia32 -w` launches the script and passes the arguments after `--`, so it builds the Windows targets in 32-bit, regardless of the OS you are running. I will document it.

Issue #244, opened 2017-10-29
When reordering tasks into a new suite, 2 tasks may have the same name


angrykoala:
By design, 2 tasks in the same suite should **not** have the same name. If the user tries to create a new task with the same name, a number (`(2)`, `(3)`....) will be appended. To do this duplicated are detected after creating a new task

However, adding a new task with the same name in a different suite is valid, so moving that task to the same suite the first task is will result in 2 tasks with the same name in the same suite.

The "Duplicate name" check should be done also whenever a task is added to a suite (either a new task or a moved taks)

Trevormichael:
I got this

PR #247, opened 2017-10-31
Issue #244 change task name to valid name after moving to new suite

Trevormichael:
## Description of the PR
As per #244, I've added a callback for the draggable add event, which is only called when moving tasks from one suite to another. The moved task is renamed using the same naming logic as when they are added. 

### Issues Related
#244

------

**Checklist**
- [ ] Your changes have tests

### DIFF
--- a/app/renderer/components/task_suite.js
+++ b/app/renderer/components/task_suite.js
@@ -37,6 +37,7 @@ module.exports = {
             data-collapsible="accordion"
             v-model="suite.tasks"
             @start="onDragStart"
+            @add="onAdd"
             :move="checkMove">
                 <template v-for="(task,i) in suite.tasks">
                     <task-card v-bind:task="task" v-on:remove="removeTask(i)" @edit="editTask(i, $event)" v-bind:event="event"></task-card>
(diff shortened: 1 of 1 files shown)


app/renderer/components/task_suite.js
```diff
@@ -66,6 +67,10 @@ module.exports = {
         onDragStart() {
             this.event.emit("collapseTask");
         },
+        onAdd(evt){
```
angrykoala: I think onAdd may be a misleading name (just below there is an `addTask` method completely unrelated)

Maybe `onDragAdd` `onTaskMovedEnd` or something like this would be better? It would be confusing to have an onAdd method that it is not called when adding a task.

On the other hand, because this logic is common to the add process and the onMoveEnd process, it could be useful to have the same method for both

app/renderer/components/task_suite.js
```diff
@@ -66,6 +67,10 @@ module.exports = {
...
+            let task = this.suite.tasks[evt.newIndex]
```
angrykoala: this should be a `const`

Issue #216, opened 2017-10-25
Tooltip with task status when hovering on status icon


angrykoala:
When hovering on the status icon on a task card, a tooltip could appear to make easier to understand its meaning:
* idle (black)
* running (blue)
* successful (green)
* failed (red)

rkomiyama:
I'd like to be assigned this issue.

angrykoala:
Consider it yours @rkomiyama 
Thanks taking interest in this project

rkomiyama:
I have the implementation done. I believe all I have left is creating test case(s).

@angrykoala Do you have a preference as to where the test case should be? I was thinking about putting it in `task.test.js`.

angrykoala:
snapshots?
eed 
Mmm, it depends on how are you focusing your tests. If the tests are only testin the Javascript part you can do them just the same way we do all the tests. If you want to test the html/css result you may need to use some tool such as Phantomjs, which we want to integrate as defined in #105

PR #252, opened 2017-10-31
Task status hover tooltip

rkomiyama:
## Description of the PR
Added functionality for tooltip when hovering over the task status icon.

### Issues Related
#216

### DIFF
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -41,7 +41,7 @@ module.exports = {
             <div class="col s1">
                 <progress-spinner v-if="running && AppStatus.config.animatedSpinner"></progress-spinner>
                 <i v-else class="small material-icons" v-bind:style="{color: statusColor}">{{task.status}}</i>
-                <tooltip v-bind:task="task"></tooltip>
+                <tooltip v-bind:taskStatus="task.status"></tooltip>
             </div>
         </div>
 
--- a/app/renderer/components/task_card.js
+++ b/app/renderer/components/task_card.js
@@ -6,8 +6,9 @@ const TaskStatus = require('../../common/task_status');
 const ProgressSpinner = require('./progress_spinner');
 const ToolTip = require('./tooltip');
 
-const Material = require('../materialize');
+const DeleteConfirmationAlert = require('../api/app_alerts').DeleteConfirmationAlert;
 const Utils = require('../../common/utils');
+const Materialize = require('../api/materialize');
 
 module.exports = {
     props: ['task', 'event'],
@@ -22,13 +23,16 @@ module.exports = {
         "tooltip": ToolTip
     },
     template: `
-    <li class="run-card">
-        <div class="collapsible-header row unselectable-text">
-            <div class="col s5">
+    <li class="run-card task-card">
+        <div class="collapsible-header row unselectable-text" v-bind:class="{ 'edit-mode': AppStatus.editMode}">
+            <div class="col s1" v-if="AppStatus.editMode">
+                <i class="tiny material-icons">drag_handle</i>
+            </div>
+            <div class="col" v-bind:class="{ s4: AppStatus.editMode, s5: !AppStatus.editMode }">
                 <strong class="truncate">{{task.title}}</strong>
             </div>
             <div class="col s3">
-                <div class="truncate task-time">{{executionTime}}</div>
+                <div class="truncate task-time" v-if="AppStatus.config.showTimer">{{executionTime}}</div>
             </div>
             <div class="col s3">
                 <a v-if="AppStatus.editMode" class="waves-effect waves-light btn delete-button" v-on:click="onDeleteClick">Delete</a>
@@ -54,6 +58,7 @@ module.exports = {
     mounted() {
         this.event.on("run", this.run);
         this.event.on("stop", this.stop);
+        this.event.on("collapseTask", this.collapseTask);
     },
     beforeDestroy() {
         this.removeListeners();
@@ -67,11 +72,13 @@ module.exports = {
         onDeleteClick(ev) {
             ev.stopPropagation();
             this.deleteTask();
-
         },
         deleteTask() {
-            this.stop();
-            this.$emit('remove');
+            const confirmationAlert = new DeleteConfirmationAlert("You will not be able to recover this task after deletion!");
+            confirmationAlert.toggle().then(() => {
+                this.stop();
+                this.$emit('remove');
+            }, () => {});
         },
         saveTask(task) {
             this.stop();
@@ -100,11 +107,7 @@ module.exports = {
             }
         },
         collapseTask() {
-            const elements = this.$el.getElementsByClassName('collapsible-header');
-            if (elements[0]) {
-                elements[0].classList.remove("active");
-                Material.updateCollapsible();
-            }
+            Materialize.collapseHeader(this.$el);
         }
     },
     computed: {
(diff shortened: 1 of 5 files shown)


test/task_card.test.js
```diff
@@ -1,7 +1,7 @@
 const Application = require('spectron').Application;
 const assert = require('assert');
 
-describe('Application launch', function () {
+describe('Spectron tests', function () {
```
angrykoala: This shouldn't have the name of the framework we are using, but describe what are we testing

app/renderer/components/tooltip.js
```diff
@@ -0,0 +1,29 @@
...
+    template: `
+    <div class="tooltip">
+        <div class="tooltip-text">
+            {{ status }}
+        </div>
+    </div>
+    `,
+    computed: {
+        status() {
+            switch (this.color) {
```
angrykoala: We shouldn't be using a prop called "color" for something that is not defining a color.

For this, we should be using the enumeration defined app/common/task_status
For example:
```js
case TaskStatus.idle:
     return "Idle" 
```

Instead of the prop "color" we should send "status" to the tooltip component

app/renderer/components/tooltip.js
```diff
@@ -0,0 +1,29 @@
...
+                    return "Stopped";
+                case "red":
+                    return "Error";
+                case "blue":
+                    return "Running";
+                case "green":
+                    return "OK";
+                case "grey":
+                default:
+                    return "N/A";
```
angrykoala: In the default case, the tooltip shouldn't be shown. There is no grey color afaik

resources/styles.css
```diff
@@ -73,12 +106,6 @@
     cursor: default;
 }
 
-.open a.edit-button.tap-target-origin, .open a.edit-button.tap-target-origin:hover {
```
angrykoala: Why is this removed? I assume is old code, but I'm not sure what was this for

resources/styles.css
```diff
@@ -147,7 +174,6 @@ input.tab-text {
 
 .unselectable-text {
     -webkit-user-select: none;
-    cursor: default;
```
angrykoala: why is this removed?

resources/styles.css
```diff
@@ -250,7 +280,6 @@ main {
     bottom: 56px;
     left: 0;
     right: 0;
-    overflow: auto;
```
angrykoala: Removing this breaks the behavour of the bottom bar.

Please, make sure you properly test any change made to the css

test/task_card.test.js
```diff
@@ -23,4 +23,19 @@ describe('Application launch', function () {
             assert.equal(count, 1);
         });
     });
+
+    it('shows tooltip on mouse hover', function () {
+        let selector = '.task-card:nth-child(1) .tooltip';
```
angrykoala: this should be a const

test/task_card.test.js
```diff
@@ -23,4 +23,19 @@ describe('Application launch', function () {
             assert.equal(count, 1);
         });
     });
+
+    it('shows tooltip on mouse hover', function () {
```
angrykoala: please, use arrow function ( `()=>{`) for anonymous function

test/task_card.test.js
```diff
@@ -1,7 +1,7 @@
 const Application = require('spectron').Application;
 const assert = require('assert');
 
-describe('Application launch', function () {
+describe('Spectron tests', function () {
     this.timeout(10000);
 
     beforeEach(function () {
```
angrykoala: Use `()=>{` here as well

test/task_card.test.js
```diff
@@ -1,7 +1,7 @@
 const Application = require('spectron').Application;
 const assert = require('assert');
 
-describe('Application launch', function () {
+describe('Spectron tests', function () {
     this.timeout(10000);
```
angrykoala: Is this timeout required?

resources/styles.css
```diff
@@ -53,6 +53,39 @@
...
+  content: '';
+  position: absolute;
+  top: 100%;
+  right: 10px;
+  border-width: 5px;
+  border-style: solid;
+  border-color: black transparent transparent transparent;
+}
+
+.tooltip:hover .tooltip-text {
```
angrykoala: Have you tried using a tooltip defined in materialize to improve the behavior of opening and closing? (some kind of animation or something?)

resources/styles.css
rkomiyama: Not sure. Must have happened when I merged or something.

resources/styles.css
rkomiyama: screen shot 2017-11-01 at 10 32 10 am

If `main` is set to `overflow: auto`, the top bar overlaps the tooltip.

test/task_card.test.js
rkomiyama: I don't think so. Removing it should be okay.

rkomiyama:
@angrykoala I've addressed your comments, but we still need to fix the tooltip overlap issue.

test/task_card.test.js
rkomiyama: Actually, this might be necessary for Travis.

test/task_card.test.js
rkomiyama: I can do that, but Mocha throws an error when I try to use arrow function for `describe()`.

angrykoala: Every other `describe` block in gaucho is using arrow function, so I don't understand why mocha would fail

app/renderer/components/task_card.js
```diff
@@ -39,17 +41,18 @@ module.exports = {
             <div class="col s1">
                 <progress-spinner v-if="running && AppStatus.config.animatedSpinner"></progress-spinner>
                 <i v-else class="small material-icons" v-bind:style="{color: statusColor}">{{task.status}}</i>
+                <tooltip v-bind:task="task"></tooltip>
```
angrykoala: This is better, but probably binding task.status will make the code in Tooltip easier and less coupled with Task

app/renderer/components/tooltip.js
```diff
@@ -0,0 +1,31 @@
...
+            {{ status }}
+        </div>
+    </div>
+    `,
+    computed: {
+        status() {
+            switch (this.task.status) {
+                case TaskStatus.idle:
+                case TaskStatus.stopped:
+                    return "Stopped";
```
angrykoala: Wich is better as text, `stopped` or `idle`? (open to discussion)

app/renderer/components/tooltip.js
```diff
@@ -0,0 +1,31 @@
...
+                case TaskStatus.error:
+                    return "Error";
+                case TaskStatus.running:
+                    return "Running";
+                case TaskStatus.ok:
+                    return "OK";
+                default:
+                    return this.task.status;
```
angrykoala: good enough

resources/styles.css
angrykoala: please check what is this for

angrykoala:
screenshot from 2017-11-01 23-40-27
As an example, your changes made the scroll bar always visible

app/renderer/components/tooltip.js
rkomiyama: I feel like `stopped` is better since `idle` makes it sound like it's suspended.

resources/styles.css
rkomiyama: I really have no idea, so I'll just put it back in.

rkomiyama:
Please let me know exactly which bugs exists. Some of them do not show up on my machine.

angrykoala:
Those where bugs related to the changes in the css you made. your last changes fixed them for what I can see.

I'll check the css carefully and let you know, overall the code looks good
Thanks

PR #256, opened 2017-11-02
Fixed dependency issues with Sortable.js

rkomiyama:
## Description of the PR
Changed `vuedraggable`'s `Sortable` dependency to the `passive` branch in order to get rid of the warning messages regarding `Sortable`'s `captureMode`.

### Issues Related
#196

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "material-design-icons-iconfont": "^3.0.3",
     "materialize-css": "^0.99.0",
     "mousetrap": "^1.6.1",
+    "sortablejs": "RubaXa/Sortable#passive",
     "sweetalert2": "^6.11.1",
     "vue": "^2.0.8",
     "vuedraggable": "^2.14.1",


angrykoala:
Hi @rkomiyama , Could you explain what is the passive branch differences?

Is it worth it using a non-master branch to fix these warnings?

rkomiyama:
@angrykoala You can see the changes [here](https://github.com/RubaXa/Sortable/commit/811e448711a63fd88099345d307be34d5b6b68c1).

Is it worth using a non-master branch? Perhaps. Although now it seems like there's a code coverage issue.

angrykoala:
Personally, if it is known to be something that will/should be fixed in the parent project, I wouldn't bother using a branch instead of the npm package. It is just a warning in the end.

If, as you say, the bug comes from the sortable library, and it is in the process of being fixed, I think is enough to document that in the issue and we can close it.

Issue #261, opened 2017-11-02
Menu should be tabable


angrykoala:
The dropdown menu tabs only work sometimes

rkomiyama:
Could you demonstrate the issue using LICEcap? LICEcap allows you to record your screen and save it as a GIF so you can post it in the GitHub comments. I'm just not entirely sure what the issue is.

angrykoala:
Steps to reproduce:
1. Open gaucho
2. open the side menu
3. Hit tab
You'll see the configuration menu highlighted to select, the others won't highlight hitting tab more times.

The desired behavior is that hitting tab once will highlight the first item, hitting again will highlight the second and so on.

Also the highlighted style may be improved instead of the yellow outline

rkomiyama:
I'm just confused as to why this functionality needs to be put in place when the side menu is not tabbable to begin with.

angrykoala:
Well, the idea is to make it tabbable

rkomiyama:
Okay, I'll look into making the side menu tabbable.

angrykoala:
Thanks

PR #262, opened 2017-11-06
Side menu and its items are now tabbable

rkomiyama:
## Description of the PR
Allow side menu to be tabble, along with its items.
tabbing around

### Issues Related
#261


app/renderer/components/navbar_menu.js
```diff
@@ -10,12 +10,12 @@ module.exports = {
     template: `
     <ul id='navbar-menu' class='dropdown-content'>
         <template v-if="!editMode">
-            <li class="unselectable-text menu-button"><a v-on:click="selected($event,'run-suite')">Run Suite</a></li>
-            <li class="unselectable-text menu-button"><a v-on:click="selected($event,'stop-suite')">Stop Suite</a></li>
+            <li class="unselectable-text menu-button"><a v-on:click="selected($event,'run-suite')" v-bind:href="'#tab0'">Run Suite</a></li>
```
angrykoala: Do we really need `v-bind:href="'#tab`? I think there are easier native ways to achieve tabbing

test/front-end.test.js
```diff
@@ -0,0 +1,83 @@
+const Application = require('spectron').Application;
+const assert = require('assert');
+
+describe('Front-end tests', function () {
+    this.timeout(20000); // needed for Travis builds
+
+    beforeEach(() => {
+        this.app = new Application({
```
angrykoala: I think using `this` in a test case is a bit obscure, Could you please create a `const` variable `app` in the describe and use that?

test/front-end.test.js
```diff
@@ -0,0 +1,83 @@
+const Application = require('spectron').Application;
+const assert = require('assert');
+
+describe('Front-end tests', function () {
```
angrykoala: Please, use arrow function here

test/front-end.test.js
```diff
@@ -0,0 +1,83 @@
...
+            })
+            .then((element) => {
+                return this.app.client.elementIdCssProperty(element.value.ELEMENT, 'visibility');
+            })
+            .then((visibility) => {
+                assert.equal(visibility.value, "visible");
+            });
+    });
+
+    it('Allows selection of side menu and its items with "Tab" key', () => {
```
angrykoala: Nice making this test! thanks

test/front-end.test.js
rkomiyama: Normally I would, but I won't be able to use `this.timeout()` if I use the arrow function here.

test/front-end.test.js
rkomiyama: Could you show me an example of what that would look like?

app/renderer/components/navbar_menu.js
rkomiyama: A simple `href` might work.

test/front-end.test.js
angrykoala: You are right, we should use `function()` on the `describe` block on the other tests

app/renderer/components/navbar_menu.js
angrykoala: I'm not sure you even need href, I think there is a tabable syntax or something like this

test/front-end.test.js
angrykoala:
```
describe("...",function(){
let app; //Sorry, it should be let, not const
beforeEach(()=>{
    app=new Application
    ....
})
```

This way, you explicitly know what variables are using, and the internals of the suite this won't collide

Issue #265, opened 2017-11-22
Add New Task Missing After 8 Tasks


bgordley:
When creating more than 8 tasks in a suite, the "Add New Task" button no longer appears.

I am using Windows 10, Gaucho version 0.4

As you can see below, I am in edit mode and do not see the "Add New Task" button.
![image](https://user-images.githubusercontent.com/7117601/33145289-b4515a40-cf74-11e7-8639-a48ef1476da9.png)

angrykoala:
Yeah, this is intentional, as 10 is the maximum amount of tasks per suite set (maybe I should add that in the docs)

Do you really need to add more tasks to a single suite? It may be worth a discussion about adding a configurable limit or removing it completely

bryangordley:
Hello @angrykoala,

My current use case has many steps as I plan to use this as a production utility. Out of curiosity, what would be the purpose of capping the maximum amount of tasks in a suite?

angrykoala:
Too many tasks may have performance and UX issues (having to scroll through the tasks). Initially this was a simple cap as Gaucho was designed to only have 4 or 5 tasks per view (based on my own use) so I decided to add a limit. Being 8 is just and arbitrary limit. In fact, the limit of suites and tasks is defined [here](https://github.com/angrykoala/gaucho/blob/master/app/renderer/app_status.js) as it was intended to be configurable.

I'll try to increase the limit or make it configurable as soon as I can, but I'm happy to accept PRs regarding this issue as well

bryangordley:
Absolutely makes sense. Thank you for the quick response! My team is enjoying this tool quite a bit.

angrykoala:
No problem, I'll figure out some improvement on this area. How many tasks do you reckon you'll be using? (to have an idea about a better upper limit)

bryangordley:
I was thinking around 15 - 20 max. Take a look at my fork if you have some time. I added a range component that has min/max controlled in app_setup. The actual values are moved to the config. I am not sure if this is the best approach or not.

angrykoala:
Please, when your feature is ready for review, do a PR so I can easily review it and point it suggestions or improvements

PR #267, opened 2017-11-24
Added ability to control amount of suites / tasks within limited range

bryangordley:
## Description of the PR
This feature grants users the ability to control the maximum amount of suites and tasks per suite within a controlled range. The original limit was 6 for suites and 8 for tasks which remain the default in this feature, maximums stopping at 12 for suites and 20 for tasks.

Additionally, made some changes to the front-end testing / package.json to support testing in a Windows environment. This will need to be tested in Mac / Linux environments. As for tests, it appears that the whole configuration menu needs tests written which I am happy to add in another pull request (or this one if required).

### Issues Related
[Add New Task Missing After 8 Tasks](https://github.com/angrykoala/gaucho/issues/265)

------

**Checklist**
- [x] All the tests and linters are passing (`npm run test`)
- [ ] Your changes have tests

### DIFF
--- a/app/common/app_config.js
+++ b/app/common/app_config.js
@@ -59,9 +59,11 @@ class AppConfig {
             encryptionKey: key
         });
     }
+
     set(key, value) {
         this.store.set(key, value);
     }
+
     get(key) {
         const result = this.store.get(key);
         return result;
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "main.js",
   "dependencies": {
     "electron-store": "^1.3.0",
-    "jquery-ui": "^1.12.1",
     "jquery": "^3.1.1",
+    "jquery-ui": "^1.12.1",
     "material-design-icons-iconfont": "^3.0.3",
     "materialize-css": "^0.99.0",
     "mousetrap": "^1.6.1",
@@ -20,6 +20,7 @@
     "cross-env": "^5.0.5",
     "electron": "~1.7.8",
     "electron-builder": "^19.41.1",
+    "electron-prebuilt": "^1.4.13",
     "eslint": "^4.8.0",
     "eslint-plugin-vue": "^3.13.1",
     "istanbul": "^0.4.5",
@@ -33,7 +34,7 @@
     "icon": "resources/icon.png"
   },
   "scripts": {
-    "test": "istanbul cover _mocha -- --exit && npm run lint",
+    "test": "istanbul cover node_modules/mocha/bin/_mocha -- --exit && npm run lint",
     "start": "electron .",
     "start-dev": "cross-env NODE_ENV=dev npm start",
     "clean": "rm -rf config.json tasks.json coverage dist",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ==================
 
   * It is possible to navigate the menu using tab
+  * Fixed bug where stderr was not being displayed
 
 0.4.0 / 2017-10-02
 ==================
--- a/app/common/suite.js
+++ b/app/common/suite.js
@@ -5,17 +5,21 @@ class Suite {
         this.title = title || "";
         this.tasks = [];
     }
+
     get length() {
         return this.tasks.length;
     }
+
     addTask(task) {
         const title = this.getValidName(task.title);
         task.title = title;
         this.tasks.push(task);
     }
+
     removeTask(index) {
         this.tasks.splice(index, 1);
     }
+
     replaceTask(index, task) {
         if (this.tasks[index].title !== task.title) {
             const title = this.getValidName(task.title);
@@ -23,6 +27,7 @@ class Suite {
         }
         this.tasks.splice(index, 1, task);
     }
+
     stopAll() {
         const promises = this.tasks.map((task) => {
             if (!task.isRunning()) return Promise.resolve();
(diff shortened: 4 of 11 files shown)


app/common/app_config.js
```diff
@@ -47,7 +49,7 @@ const defaultTasks = {
     }]
 };
 
-// Just for basic offuscation of the config file
+// Just for basic obfuscation of the config file
```
angrykoala: Good catch, thanks

app/renderer/components/range_form.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
```
angrykoala: I like this component, easy to re-use

package.json
```diff
@@ -33,7 +34,7 @@
     "icon": "resources/icon.png"
   },
   "scripts": {
-    "test": "istanbul cover _mocha -- --exit && npm run lint",
+    "test": "istanbul cover node_modules/mocha/bin/_mocha -- --exit && npm run lint",
```
angrykoala: I assume this is required for the mac testing

package.json
```diff
@@ -20,6 +20,7 @@
     "cross-env": "^5.0.5",
     "electron": "~1.7.8",
     "electron-builder": "^19.41.1",
+    "electron-prebuilt": "^1.4.13",
```
angrykoala: It is really required `electron-prebuilt` in order to run Spectron in windows? I would like to avoid extra dependencies if possible

angrykoala:
Hi @bryangordley 
Thanks for your PR.
I agree with the range of maximum tasks per suite you added. But I have a couple of questions:
* If a user create a suite with 10 tasks for example, and then the maximum tasks are moved to 8, now we have an inconsistent suite with more than the maximum tasks.
* What should we do when importing tasks from a configuration with a bigger maximum of tasks?

The suites present a different problem, the limit of suites was added because more than 8 suites will not be displayed correctly (as shown in the picture) and will make suite tabs extremely small as adding a scroll is not feasible. 

screenshot from 2017-11-25 16-22-39

After reviewing how it works with a configurable maximum, I feel that it doesn't make sense to make this configurable, it just over-complicates the usage.

If Gaucho works fine with 20 tasks per suite (that's 160 tasks in total!) I think it would be easier to simply make the hardcoded limit 20.

It may be useful, however, to add some "advanced options" with a numeric field to add any number of maximum tasks for people who actually need more than the maximum tasks, but that would be a separate issue. 
What do you think about all this?

I agree with the changes made for testing (check the questions in some comments). Also, most of Gaucho UI require proper e2e testing, including configuration. I'm happy to add an issue for configuration testing if you want to do it. I'll try to work on improving current tests with Spectron

package.json
bryangordley: You are correct. I am not sure if this works with Mac or Linux build environments. I will give it a test run in a fresh Linux installation tonight.

app/renderer/components/range_form.js
bryangordley: Thanks, I made my best attempt at following the structure of the existing components.

package.json
bryangordley: Unfortunately I could not find another way to resolve the issue with Spectron on Windows. What I have discovered is that './node_modules/.bin/electron' should work properly but has a bug somewhere downstream.

I will do some more research and see if there is a way around this issue without adding dependencies.

bryangordley:
Below is my stance on the points you brought to the table
- I agree that hard-coding the tasks to 20 will cover most use cases. I like the idea of an advanced section in the future that allows a configurable max in the form of a numeric input. 
- I will remove the suite range as it seems like this is not very useful and creates more issues than it resolves.
- Creating tests for the existing Configuration options would be a great learning experience. I understand how the testing environment works and am pretty comfortable with Spectron at this point.

Below are some questions that I have
- Would you like to keep the hard-coded values stored in AppStatus.Config per my change or move back into AppStatus?
- Would you like to retain the range_form.js component for future use?

angrykoala:
Hi @bryangordley 
* If 20 tasks (or maybe less if possible) are enough for your use case of Gaucho, I think is just easier to hardcode that value for now rather than adding extra complexity to the current config, which is not designed for a lot of options. Keep in mind that the limit was added simply because I'm not sure how Gaucho will handle a huge amount of tasks.
* Perfect, I'm writing some E2E tests these days, trying to clean up a bit, I think it'll make your own tests easier, thanks, I need a bit of help writing tests with Spectron
* If you like the idea of having advanced configs, please, feel free to open an issue, once we define 2 or 3 advanced features, those could be added as a sub-menu of the current config
* All hard-coded values that aren't going to be changed in the config should go back to AppStatus (which, atm is the placeholder for a proper config file and vuex storage)
* Yes, I'm not a fan of having unused components, but I think we can use it very soon so I'm happy to keep that component in the code

Thanks!

bryangordley:
@angrykoala Awesome, this is my first time contributing to a project on GitHub, I am still learning the ropes a bit. Is there a way to update this PR with the changes you have requested or should I close this PR and submit a new one?

angrykoala:
Hi @bryangordley no worries, that's what these projects are for.

Usually, just modifying your own branch should automatically update the PR

angrykoala:
Since 0.4.1, number of tasks per suites is increased, this PR will be closed as this should fix the issue discussed

